Issue #21, opened 2020-08-12
Commands output


fabriciojs:
Make sure that error output for any commands are properly returned back to original shell output.

https://blog.kowalczyk.info/article/wOYk/advanced-command-execution-in-go-with-osexec.html

fbrissi:
Check here:
https://github.com/kool-dev/kool/blob/88a89a2893c038b2fa5e41c9b98ab995c1eb94cc/cmd/start.go#L37-L41

When the docker is not up, the output is only `2020/08/12 12:09:46.573708 exit status 1`

fabriciojs:
This has been improved and we will consider fixed as of the latest release `1.0.15`.

Please open a new descriptive issue if necessary.

PR #37, opened 2020-08-26
feat(start): Adding check for docker daemon

danielsuguimoto:
Relates to #21

### DIFF
--- a/cmd/start.go
+++ b/cmd/start.go
@@ -29,18 +29,9 @@ func init() {
 }
 
 func runStart(cmd *cobra.Command, args []string) {
-	checkDaemon()
+	checkKoolDependencies()
 	handleGlobalNetwork()
 	startContainers(startFlags.Services)
-}
-
-func checkDaemon() {
-	_, err := shellExec("docker", "info")
-
-	if err != nil {
-		execError("Docker daemon doesn't seem to be running, run it first and retry.", err)
-		os.Exit(1)
-	}
 }
 
 func handleGlobalNetwork() {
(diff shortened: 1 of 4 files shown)


cmd/start.go
```diff
@@ -29,10 +29,20 @@ func init() {
 }
 
 func runStart(cmd *cobra.Command, args []string) {
+	checkDaemon()
 	handleGlobalNetwork()
 	startContainers(startFlags.Services)
 }
 
+func checkDaemon() {
```
fabriciojs, line 37: Instead of `checkDaemon`, maybe we would benefit on renaming this function to something like `checkKoolDependencies`.

cmd/start.go
```diff
@@ -29,10 +29,20 @@ func init() {
...
+	_, err := shellExec("docker", "info")
```
fabriciojs, line 38: I understand possibly the `docker info` is something that passes/fails quickly enough. Just need to double check - under Linux, if Docker Daemon is not running, will this indeed error out?

fabriciojs: Now, a second point - other than checking `docker info`, before it I would like to do two less expansive checks that are great for most use cases we will have - let's use `exec.LookPath` to confirm that both `docker` and `docker-compose` exist and are available for execution in the current PATH settings.

cmd/start.go
```diff
@@ -29,10 +29,20 @@ func init() {
 }
 
 func runStart(cmd *cobra.Command, args []string) {
+	checkDaemon()
```
fabriciojs, line 32: I think we should also call this on at least a couple other commands, without risking too much overhead: `stop` and  `status` comes to my mind.

Issue #48, opened 2020-09-03
'kool run compile' ignoring .env


danielsuguimoto:
The command that is being exceuted: `docker run --init --rm -w /app -ti --volume /Users/jun/Opensource/kool:/app golang:1.15.0 go build -o kool`

I guess that the golang docker image is Linux based and we are not forwarding the local environment variables, but not 100% sure.

Obs: my local environment is MacOS based

fabriciojs:
I am not sure I understand. What environment variables you would like to pass down the container? And .env is not something I ever considered supporting for compilation time.

Passing variables through for the container would require using `docker run -e ENV=VAL` in order to pass them down, which we don't do - and probably do not want to do for all variables.

So to clarify in the given situation:

- `docker` does have all the env environment available.
- `go build` does not have, since `docker run` did not pass down the variables with the `-e` option.

fabriciojs:
Ahh I understand now, you mean for local development.

I guess in this case you can setup the `.env.dist` file with the options (Linux / MacOS) for GOOS, GOARCH, etc. Default to Linux.

Then ignore `.env`. Then update the `compile` script within `kool.yml` to pass down the necessary variables explicitly.

fabriciojs:
This has been solved at current master and will go into the next release 1.0.17.

PR #56, opened 2020-09-06
Fix "kool run compile" by improving kool docker to parse pre-image options

fabriciojs:
https://github.com/kool-dev/kool/issues/48

### DIFF
--- a/cmd/docker.go
+++ b/cmd/docker.go
@@ -30,8 +30,7 @@ func init() {
 
 func runDocker(cmd *cobra.Command, originalArgs []string) {
 	var (
-		image string
-		args  []string
+		args []string
 	)
 
 	if originalArgs[0] == "--disable-tty" || originalArgs[0] == "-T" {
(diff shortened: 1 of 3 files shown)


dbpolito (approved):

dbpolito:
looking at this i see you fixed basically all problems we had with --env, -p, etc.

Now that we have this i somehow think we need a way to disable / skip default options from kool, like volume / workdir change... because the use might want and can change this behaviour.

Maybe an `--skip-default-options`. 🤔

fabriciojs:
> looking at this i see you fixed basically all problems we had with --env, -p, etc.
> 
> Now that we have this i somehow think we need a way to disable / skip default options from kool, like volume / workdir change... because the use might want and can change this behaviour.
> 
> Maybe an `--skip-default-options`. 🤔

I hear you... but I guess on that case it is easier for anyone to actually use `docker run` instead, right... we should clarify that `kool docker` is really indented to be a helper by given default folder/rm best for context aware containers execution.

PR #64, opened 2020-09-09
feat(envs): Env variables on docker and exec

### LINKED ISSUE #63: Support environment variables injection on kool exec
It is important to have `-e VAR=VAL` working and available to use on `kool exec`. It is a recurring need on several projects.

danielsuguimoto:
Closes #63 

Also refactored this same feature on the docker command.

### DIFF
--- a/cmd/docker.go
+++ b/cmd/docker.go
@@ -10,13 +10,13 @@ import (
 // DockerFlags holds the flags for the docker command
 type DockerFlags struct {
 	DisableTty bool
+	EnvVariables []string
 }
 
 var dockerCmd = &cobra.Command{
 	Use:                "docker [options] [image] [command]",
 	Args:               cobra.MinimumNArgs(1),
 	Run:                runDocker,
-	DisableFlagParsing: true,
 	Short:              "Creates a new container and runs the command in it.",
 	Long: `This command acts as a helper for docker run.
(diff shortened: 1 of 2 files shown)


cmd/docker.go
```diff
@@ -10,49 +10,44 @@ import (
...
 	dockerCmd.Flags().BoolVarP(&dockerFlags.DisableTty, "disable-tty", "T", false, "Disables TTY")
-}
+	dockerCmd.Flags().StringArrayVarP(&dockerFlags.EnvVariables, "env", "e", []string{}, "Environment variables")
 
-func runDocker(cmd *cobra.Command, originalArgs []string) {
-	var (
-		args []string
-	)
+	//After a non-flag arg, stop parsing flags
+	dockerCmd.Flags().SetInterspersed(false)
```
fabriciojs, line 36: Nice!

fabriciojs: So this removes the need for `DisableFlagParsing` which is nice. But in the other hand, we are only now able to set the --env option for `docker run`, and not any others like i.e `-p 80:80`. I assume that is OK for now, although it decreases the flexibility, I see more benefit on improving how we handle the flags at this moment.

fabriciojs (approved):
👍🏻

PR #66, opened 2020-09-09
fix(docker): Added more flags to docker commands

danielsuguimoto:


dbpolito (approved):

cmd/docker.go
```diff
@@ -66,7 +72,23 @@ func execDockerRun(image string, command []string) {
...
+			args = append(args, "--volume", volume)
+		}
+	}
+
+	if dockerFlags.Expose != "" {
+		args = append(args, "--expose", dockerFlags.Expose)
+	}
+
+	if dockerFlags.Publish != "" {
+		args = append(args, "--publish", dockerFlags.Publish)
```
fabriciojs, line 88: For `--publish` (short `-p`) I guess we should also make it an array like Volumes and Envs... there are common cases that we want to export two ports from a `kool docker` (i.e when serving a Node application and we want to connect DevTools to it).

cmd/docker.go
```diff
@@ -21,16 +24,19 @@ var dockerCmd = &cobra.Command{
...
-var dockerFlags = &DockerFlags{false, []string{}}
+var dockerFlags = &DockerFlags{false, []string{}, []string{}, "", ""}
 
 func init() {
 	rootCmd.AddCommand(dockerCmd)
 
 	dockerCmd.Flags().BoolVarP(&dockerFlags.DisableTty, "disable-tty", "T", false, "Disables TTY")
 	dockerCmd.Flags().StringArrayVarP(&dockerFlags.EnvVariables, "env", "e", []string{}, "Environment variables")
+	dockerCmd.Flags().StringArrayVarP(&dockerFlags.Volumes, "volume", "v", []string{}, "Bind mount a volume")
+	dockerCmd.Flags().StringVar(&dockerFlags.Expose, "expose", "", "Expose a port or a range of ports")
```
fabriciojs, line 38: I don't remember ever using the `--expose` - just wondering do we really need it?

fabriciojs (changes requested):

fabriciojs (approved):

Issue #23, opened 2020-08-12
Proposal: kool logs


fabriciojs:
`kool logs [--follow] [--tail] {service}`

Also, consider having the same API for `kool deploy logs`

fabriciojs:
This has been implemented and merged onto `master`. This will be on next release.

PR #70, opened 2020-09-09
feat(logs): New logs command

danielsuguimoto:
Relates to #23


cmd/logs.go
```diff
@@ -0,0 +1,52 @@
...
+	if logsFlags.Tail > 0 {
+		args = append(args, "--tail", strconv.Itoa(logsFlags.Tail))
+	}
+
+	if logsFlags.Follow {
+		args = append(args, "--follow")
+	}
+
+	for _, service := range originalArgs {
+		args = append(args, service)
```
fabriciojs, line 43: You can make this a one liner with `args = append(args, originalArgs...)`

fabriciojs:
@danielsuguimoto @dbpolito - technically this PR is OK, but I would like to bring up a discussion on the defaults... right now the default is only a proxy to `docker-compose logs`. I think that it would be a good  improvement to default with a tail of like 25, and defaulting to `follow`. But I have to be honest, this is more a gut feeling than actual data-backed opinion.

What do you guys think?

dbpolito:
I agree, printing everything always is a bad default...  Defaulting `tail` to 25 is a good start, then the user can change it to 0 our something else...

Good to somehow explain this behaviour on the description of the command.

dbpolito:
Also i think we can bring the argument `service` from docker-compose log, in case you want to see logs from only a service... Not sure how docker-compose handles one service or multiple services....

danielsuguimoto:
@fabriciojs @dbpolito should we allow `--tail` accept a string value like `all`? We can add a new flag `--all` also...

dbpolito:
I would have just tail and follow, tail=0 would mean all

fabriciojs:
👍🏻  for `--follow` as an option, and having `--tail=25` by default, and document on the flag _if you want all the logs set --tail=0`... For me this is usually the better route because logs from containers may be too long, and we usually are looking only the latest lines.

fabriciojs:
@danielsuguimoto just a  heads up - now that we merged the refactor of Shell to master, you need to update this branch and apply the change from `shellInteractive` to `shell.Interactive` as well.

fabriciojs (approved):

dbpolito (approved):

PR #78, opened 2020-09-10
docs: Refactoring docs generation

danielsuguimoto:
Trying to fix and improve docs generation to make them more dynamically


docs/make_docs.sh
```diff
@@ -3,17 +3,3 @@
 rm -f 4-Commands/*.md
 
 go run docs.go
```
fabriciojs, line 5: Now that this file became only two commands, I guess would be good to change remove this file and leave them only on kool.yml?

```yml
  make-docs:
    - rm -f docs/4-Commands/*.md
    - kool run go run docs/docs.go
```

PS: we might even move `docs.go` to the root folder, and then add a `+build ignore` annotation to it.

fabriciojs (commented):
Great! 👍🏻 

Just left a comment on maybe removing the bash script in favour of kool.yml only command.

fabriciojs (approved):
Good job!

Issue #20, opened 2020-08-08
Implement test suites


fabriciojs:
After some initial effort we have now a test pipeline in place, and a few examples on testing some important code paths.

---

Our current graph is not very promising though, but we will get there!

![CodeCov](https://codecov.io/gh/kool-dev/kool/branch/master/graphs/sunburst.svg)

fabriciojs:
Once #92 is merged we can check out https://codecov.io/gh/kool-dev/kool and grab ourselves a good old badge to display.

danielsuguimoto:
After merging #95 and refactoring the entire application to use the new command builder added on #93, we will be able to mock most of the commands and increase testing coverage a lot.

dbpolito:
I'm closing this one as we reached our initial goal which was 90% coverage, thanks @danielsuguimoto for your hard work on this one.

PR #92, opened 2020-09-12
POC: testing

fabriciojs:
#20


dbpolito (approved):

kool.yml
```diff
@@ -2,7 +2,6 @@ scripts:
   # Helper for local development - parsing presets onto Go,
   # compiling and installing locally
   dev:
-    - kool run parse-presets
```
dbpolito, line 5: really wanted to remove this one?

fabriciojs: yeah, people are using it to compile/install locally only under testing environment... the parse-presets is only meant when we update presets so, went for removing from `dev`

PR #95, opened 2020-09-14
tests(Checker): Adding test for dependencies checker

danielsuguimoto:
Relates to #20


cmd/checker/checker.go
```diff
@@ -0,0 +1,59 @@
...
+	var (
+		err error
+		dockerInfoCmd *builder.Command
+		dockerComposePsCmd *builder.Command
+	)
+
+	dockerInfoCmd, err = builder.ParseCommand("docker info")
+
+	if err != nil {
+		log.Fatal(err)
```
fabriciojs, line 30: I think our packages should never "fatal", but instead return an `error`. So the proper exit control is handled in a single layer at the Cobra commands I imagine.

cmd/checker/checker.go
```diff
@@ -0,0 +1,59 @@
...
+
+// NewChecker initializes checker
+func NewChecker() Checker {
+	var (
+		err error
+		dockerInfoCmd *builder.Command
+		dockerComposePsCmd *builder.Command
+	)
+
+	dockerInfoCmd, err = builder.ParseCommand("docker info")
```
fabriciojs, line 27: Since this is a piece of code we "trust", I think it is best we build the command instead of parsing it out of a string. This is because of load at compile time instead of runtime (parsing the string).

We could have:

```go
dockerInfoCmd = &builder.Command{"docker", []string{"info"}}
```

fabriciojs: So there is no error to be handled.

cmd/checker/checker.go
```diff
@@ -0,0 +1,59 @@
...
+
+	if err != nil {
+		log.Fatal(err)
+	}
+	dockerComposePsCmd, err = builder.ParseCommand("docker-compose ps")
```
fabriciojs, line 32: Same as the docker info situation above.

cmd/checker/checker.go
```diff
@@ -0,0 +1,59 @@
+package checker
+
+import (
+	"kool-dev/kool/cmd/builder"
+	"log"
+)
+
+// Checker defines the check kool dependencies method
+type Checker interface {
+	CheckKoolDependencies() (string, error)
```
fabriciojs, line 10: This interface should only return an `error`. One more comment about this below on its implementation for DefaultChecker.

cmd/checker/checker.go
```diff
@@ -0,0 +1,59 @@
...
+		log.Fatal(err)
+	}
+
+	return &DefaultChecker{dockerInfoCmd, dockerComposePsCmd}
+}
+
+// CheckKoolDependencies checks kool dependencies
+func (c *DefaultChecker) CheckKoolDependencies() (message string, err error) {
+	if err = c.DockerCmd.LookPath(); err != nil {
+		message = "Docker doesn't seem to be installed, install it first and retry."
```
fabriciojs, line 44: This `message` should be our error itself. We can do in the global scope of this package in a file `errors.go`:

```
var ErrDockerNotInstalled = errors.New("Docker doesn't seem to be installed, install it first and retry.")

...
```

Then here we just return each error we have. This is helpful for checking the error later, having them globally pre-defined.

cmd/checker/checker.go
fabriciojs: Maybe now that this moved onto its own package the name of this function could become something like `VerifyDependencies`

cmd/checker/checker_test.go
```diff
@@ -0,0 +1,143 @@
...
+	c = &DefaultChecker{dockerCmd, dockerComposeCmd}
+
+	message, err := c.CheckKoolDependencies()
+
+	if err == nil {
+		t.Error("Expected an error, got none.")
+		return
+	}
+
+	if message != "Docker doesn't seem to be installed, install it first and retry." {
```
fabriciojs, line 80: By changing the Err* handling above suggested, here we could have only helper methods like `IsDockerNotFoundError(err)` which then implements the `err.Error() == ErrDockerNotInstalled.Error()`

cmd/builder/command.go
```diff
@@ -15,6 +16,15 @@ type Command struct {
 	args    []string
 }
 
+// Executor holds available methods for command.
+type Executor interface {
```
fabriciojs, line 20: Although it is nice to have an verb as the interface name, I think in this case we could have the interface named `Command` only, and then the current struct `Command` becomes `DefaultCommand` - what do you think?

cmd/builder/command.go
```diff
@@ -15,6 +16,15 @@ type Command struct {
...
+	AppendArgs(args ...string)
+	Interactive() error
+	String() string
+	Exec() (string, error)
```
fabriciojs, line 24: This note should be further discussed and dealt with down the road (next PR maybe) but here it goes - I do think we should split the logic of command builder and runner.

Two different tasks, which could make us have in the end a `Builder` interface and `DefaultBuilder` struct for the commands building, and then a `Runner` `DefaultRun` which encapsulates `shell.Exec/shell.Interactive` etc.

cmd/builder/command.go
```diff
@@ -41,7 +51,19 @@ func (c *Command) Interactive() (err error) {
...
+}
+
 // String returns a string representation of the command.
 func (c *Command) String() string {
 	return strings.Trim(fmt.Sprintf("%s %s", c.command, strings.Join(c.args, " ")), " ")
 }
+
+// LookPath returns if the command exists
+func (c *Command) LookPath() (err error) {
+	_, err = exec.LookPath(c.command)
```
fabriciojs, line 67: Scratching my head thinking on how  to test this. Well at least we could have a dummy test with an unknown executable and verify the error is returned.

To mock the LookPath return then I am not sure (non-error return); have that in a safe way... maybe checking for `go` command? (assuming tests are executed in an environment that MUST have `go` installed, right)

cmd/start.go
```diff
@@ -30,7 +31,14 @@ func init() {
 }
 
 func runStart(cmd *cobra.Command, args []string) {
-	checkKoolDependencies()
+	var dependenciesChecker = checker.NewChecker()
+	message, err := dependenciesChecker.CheckKoolDependencies()
+
+	if (err != nil) {
```
fabriciojs, line 37: please run `kool run fmt` - it will remove this parenthesis I believe hehe

fabriciojs (changes requested):
@danielsuguimoto nice job!

I added a few comments on some topics...

beyond those comments, I would like to suggest a thinking on how we test indeed the command execution...

this here sounded clever although hacky - https://jamiethompson.me/posts/Unit-Testing-Exec-Command-In-Golang/

cmd/builder/command.go
danielsuguimoto: @fabriciojs yes, we need to think better in an easy and fashionable way to mock both `exec.LookPath` and `exec.Command`. Make them easily replaceable for testing purposes.

fabriciojs (approved):

PR #102, opened 2020-09-18
tests(status): Refactoring status command for testing

danielsuguimoto:
Relates to #20

### DIFF
--- a/cmd/builder/command.go
+++ b/cmd/builder/command.go
@@ -24,8 +24,8 @@ type Builder interface {
 
 // Runner holds available methods for running commands.
 type Runner interface {
-	Interactive() error
-	Exec() (string, error)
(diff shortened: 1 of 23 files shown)


fabriciojs (commented):
So far, I like it!

There are a couple of points I would like to think about doing some other fashion, but from where we are coming to this is already awesome!

To note what I would want us to think through in future iterations:

- Error/Output... Maybe we should create an Output handler which encapsulates the io.Writer, then we always us it so we don't need to pass io.Writer so many times from so many places.

- Command interface, like StatusCmd, either that should be made a package or simplify it (smaller/encapsulate more behaviour).

danielsuguimoto:
Closing in favor of #103

PR #115, opened 2020-09-24
tests(run): adding tests to run command

danielsuguimoto:
Relates to #20

### DIFF
--- a/cmd/builder/fake_command.go
+++ b/cmd/builder/fake_command.go
@@ -52,16 +52,14 @@ func (f *FakeCommand) Exec(args ...string) (outStr string, err error) {
 
 // Interactive will send the command to an interactive execution.
 func (f *FakeFailedCommand) Interactive(args ...string) (err error) {
-	f.CalledInteractive = true
-	f.ArgsInteractive = args
+	_ = f.FakeCommand.Interactive(args...)
 	err = f.MockError
 	return
 }
 
 // Exec will send the command to shell execution.
 func (f *FakeFailedCommand) Exec(args ...string) (outStr string, err error) {
-	f.CalledExec = true
-	f.ArgsExec = args
+	_, _ = f.FakeCommand.Exec(args...)
 	err = f.MockError
 	return
 }
--- a/cmd/builder/fake_command.go
+++ b/cmd/builder/fake_command.go
@@ -12,6 +12,12 @@ type FakeCommand struct {
 	CalledExec        bool
 }
 
+// FakeFailedCommand implements the Command interface and is used for mocking failing testing scenarios
+type FakeFailedCommand struct {
+	FakeCommand
+	MockError error
+}
+
 // AppendArgs mocked function for testing
 func (f *FakeCommand) AppendArgs(args ...string) {
 	f.ArgsAppend = append(f.ArgsAppend, args...)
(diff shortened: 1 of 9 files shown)


cmd/builder/fake_command.go
```diff
@@ -43,3 +49,19 @@ func (f *FakeCommand) Exec(args ...string) (outStr string, err error) {
 	f.ArgsExec = args
 	return
 }
+
+// Interactive will send the command to an interactive execution.
+func (f *FakeFailedCommand) Interactive(args ...string) (err error) {
+	f.CalledInteractive = true
```
fabriciojs, line 55: here we can reuse the logic already implemented: `f.FakeCommand.Interactive(args...)` so we don't repeat it!

cmd/builder/fake_command.go
```diff
@@ -43,3 +49,19 @@ func (f *FakeCommand) Exec(args ...string) (outStr string, err error) {
...
+	f.ArgsInteractive = args
+	err = f.MockError
+	return
+}
+
+// Exec will send the command to shell execution.
+func (f *FakeFailedCommand) Exec(args ...string) (outStr string, err error) {
+	f.CalledExec = true
```
fabriciojs, line 63: same thing `f.FakeCommand.Exec(args...)` instead of duplicating the two lines

cmd/builder/fake_command_test.go
```diff
@@ -35,3 +38,41 @@ func TestFakeCommand(t *testing.T) {
 		t.Errorf("failed to use mocked Exec function on FakeCommand")
 	}
 }
+
+func TestFakeFailedCommand(t *testing.T) {
+	f := &FakeFailedCommand{MockError: errors.New("error")}
+
+	f.AppendArgs("arg1", "arg2")
+
+	if !f.CalledAppendArgs || f.ArgsAppend == nil || f.ArgsAppend[0] != "arg1" || f.ArgsAppend[1] != "arg2" {
```
fabriciojs, line 47: we do not override the AppendArgs, so I don't think we need to test it again on FakeFailedCommand

cmd/builder/fake_command_test.go
```diff
@@ -35,3 +38,41 @@ func TestFakeCommand(t *testing.T) {
...
+		t.Errorf("failed to use mocked AppendArgs function on FakeFailedCommand")
+	}
+
+	_ = f.String()
+
+	if !f.CalledString {
+		t.Errorf("failed to use mocked String function on FakeFailedCommand")
+	}
+
+	_ = f.LookPath()
```
fabriciojs, line 57: same, shouldn't need to test

fabriciojs (changes requested):
Que coisa mais linda! 😃 

Awesome! Just made a few comments on struct composing, so we re-use the original methods without re-implementing.

fabriciojs (approved):
👏🏻

fabriciojs:
@danielsuguimoto for another testing iteration, still related to `kool run` - we have the `presets map[string]map[string]string` variable which gets its content from the presets/ folder... I see as helpful if we add in a test suite that iterates over the presets map looking for `kool.yml` files, and once finding one, asserts that such YML is valid (can be properly parsed by our parser).

PR #116, opened 2020-09-24
tests(presets): adding presets test

danielsuguimoto:
Relates to #20


fabriciojs (approved):
Awesome! Now we can trust when people add in/make changes to presets, that it won't break only at run time after someone tries using it!

PR #120, opened 2020-09-28
tests(docker): adding tests to docker command

danielsuguimoto:
Relates to #20

### DIFF
--- a/cmd/docker.go
+++ b/cmd/docker.go
@@ -1,7 +1,7 @@
 package cmd
 
 import (
-	"kool-dev/kool/cmd/shell"
+	"kool-dev/kool/cmd/builder"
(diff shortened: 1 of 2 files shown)


fabriciojs (approved):
This looks good!

Just FYI @danielsuguimoto - on `kool docker` recently I had to work with more pre-image flags than what we currently support here...

in order to do so I did:

`kool docker -- --network=xxx -p 80:80 image:tag ...`

I believe Cobra handles the `--` sign as `stop parsing args` - that way, after that point, we can provide any options as an usual argument... given this behaviour, I think we should drop the proxy we currently implement for volumes/env, and have this `--` thing properly documented on the command long description. We can handle that soon in another PR.

PR #134, opened 2020-10-01
tests(restart): improving restart command tests

danielsuguimoto:
Relates to #20


fabriciojs (approved):
@danielsuguimoto this looks good! Just one thought... we have to do stuff like this for all default KoolService / cobra.Command we create and the `Run: func (...` have that default sequence - setOut; Exec; if err, Errror(err), etc...

Maybe we could turn this func into a known/repeatable function implementation, so we test it only once, and reuse everywhere. Just one idea for a possible future improvement...

PR #138, opened 2020-10-02
refactor(command): new way of creating command

danielsuguimoto:
It's just a proposal of a new way to create commands, avoiding repeating some code, for example, the command `Run`function. If approved, I can refactor the other commands, but before I want to know what do you guys think about it.


cmd/root.go
```diff
@@ -25,3 +33,27 @@ func Execute() error {
 func RootCmd() *cobra.Command {
 	return rootCmd
 }
+
+// CreateCommand creates a new command
+func CreateCommand(service KoolService, values NewCommandData) *cobra.Command {
```
fabriciojs, line 38: func CreateCommand(service KoolService, values NewCommandData) (newCmd *cobra.Command) {

fabriciojs: we will have different needs for other flags/parameters depending on the command... which makes me think that maybe `NewCommandData` is not worth.

fabriciojs:
@danielsuguimoto I am not sure I dig this initial proposal. Maybe if we simply looked at returning a `cobraRunFn` given a `KoolService` would be simpler and clearer.

danielsuguimoto:
@fabriciojs you mean, a `DefaultCommandRunFunction(service KoolService)` that returns the run function itself?

fabriciojs:
> @fabriciojs you mean, a `DefaultCommandRunFunction(service KoolService)` that returns the run function itself?

Yes!

cmd/root.go
```diff
@@ -25,3 +28,15 @@ func Execute() error {
 func RootCmd() *cobra.Command {
 	return rootCmd
 }
+
+// DefaultCommandRunFunction default run function logic
+func DefaultCommandRunFunction(service KoolService) CobraRunFN {
```
fabriciojs, line 33: This is it! nice.

Now, I am not sure how many of our commands will get this DefaultCommandRunFunction, but maybe one of them already have a special situation (the `restart`) where we have a list of KoolService... so maybe this could already be generic, receiving `DefaultCommandRunFunction(service KoolService...)`, and then we do the same execution thing, one by one, in a loop, failing in the first error.

fabriciojs: Also, a great thing for this is, we should write a test for this generic function based on KoolService, passing FakeKoolServices, and then we test this whole thing only once! and tests for these default commands become a bit less redundant/repetitive.

fabriciojs (changes requested):
I like it!

fabriciojs:
@danielsuguimoto I like this proposal - what do you think? 

cc @dbpolito

dbpolito:
Looks good to me

fabriciojs (approved):
I really like how this ended up! Good job @danielsuguimoto !

PR #141, opened 2020-10-05
feat(tty): adding auto tty config

### LINKED ISSUE #137: Automatically tell TTY environment so we don't need flags/env vars
**Describe the feature**

Would be great if we could automatically tell whether we are running in a terminal of not, so toggle the TTY flags on Docker and Docker Compose commands.

No BC concerns.

**Use cases**

Running commands in a Cron or CI environment, we would no longer explicitly disable TTY.

**Resources**

- https://rosettacode.org/wiki/Check_output_device_is_a_terminal#Go

danielsuguimoto:
Closes #137


cmd/docker.go
```diff
@@ -48,7 +51,7 @@ func (d *KoolDocker) Execute(args []string) (err error) {
 	image := args[0]
 	workDir, _ := os.Getwd()
 
-	if !d.Flags.DisableTty && !environment.IsTrue("KOOL_TTY_DISABLE") {
+	if !d.Flags.DisableTty && !environment.IsTrue("KOOL_TTY_DISABLE") && d.terminal.IsTerminal(d.GetWriter()) {
```
fabriciojs, line 54: so, with this change we want to remove (deprecate) the flag and the environment variable, so they are not needed anymore.

fabriciojs: I guess we can do that, and target this to `2.0.0`.

cmd/shell/terminal.go
```diff
@@ -0,0 +1,22 @@
...
+// DefaultTerminalChecker holds data to check if environment is a terminal
+type DefaultTerminalChecker struct{}
+
+// IsTerminal checks if input is a terminal
+func (t *DefaultTerminalChecker) IsTerminal(in interface{}) (isTerminal bool) {
+	_, isTerminal = term.GetFdInfo(in)
+	return
+}
+
+// NewTerminalChecker creates a new output writer
```
fabriciojs, line 19: `... creates a new terminal checker`

cmd/shell/terminal.go
```diff
@@ -0,0 +1,22 @@
...
+type TerminalChecker interface {
+	IsTerminal(interface{}) bool
+}
+
+// DefaultTerminalChecker holds data to check if environment is a terminal
+type DefaultTerminalChecker struct{}
+
+// IsTerminal checks if input is a terminal
+func (t *DefaultTerminalChecker) IsTerminal(in interface{}) (isTerminal bool) {
+	_, isTerminal = term.GetFdInfo(in)
```
fabriciojs, line 15: I think we need to add a `terminal_test.go` file where we write two tests - confirming we are or not in a TTY environment.

On one test, create a pipe, send the writer piece of it down for checking, and asserts it is NOT a valid tty.

On the other test, we create something with PTY (https://github.com/creack/pty):

- We start something (anything - a real command).

```go
        // Create arbitrary command.
        c := exec.Command("bash")

        // Start the command with a pty.
        ptmx, err := pty.Start(c)

       // now we test ptmx (*os.File) for being or not a TTY!
```

https://github.com/creack/pty/blob/a6c0a376f1d083f0ced41ae7bb811497b432f3cd/run.go#L44-L74

- We check the returned `tty` passes the moby check for `isTerminal`.

fabriciojs (changes requested):
Nice work! The tests, and the breaking change, I believe are worth, for making 2.0! we can use the opportunity for more API adjustments.

cmd/shell/terminal.go
danielsuguimoto: oops

cmd/docker.go
danielsuguimoto: @fabriciojs just to confirm: deprecate or remove them?

fabriciojs: Remove!

fabriciojs: well, actually, what do you think? I would say remove, but that requires us to go to `2.0.0`... if we just deprecate, we keep the flag - change its documentation to DEPRECATED with NO BEHAVIOUR attached to it, and then we can keep version `1.3.x`... maybe that is better short term?

danielsuguimoto: @fabriciojs I have no idea if anyone is using these flags. But when in doubt, I choose the safest way: deprecating.

fabriciojs: Let's do with deprecation, and target removal at 2.0 👍🏻

danielsuguimoto: @fabriciojs Nice! Warn about deprecated stuff on documentation is enough, or should we output a message when deprecated features are used as well?

cmd/shell/terminal_test.go
```diff
@@ -0,0 +1,51 @@
+// +build !windows
```
danielsuguimoto, line 1: github.com/creack/pty has no windows support =\

cmd/docker.go
fabriciojs: I am good with only documentation, which we already get by updating the flag description.

cmd/docker.go
```diff
@@ -48,7 +51,15 @@ func (d *KoolDocker) Execute(args []string) (err error) {
 	image := args[0]
 	workDir, _ := os.Getwd()
 
-	if !d.Flags.DisableTty && !environment.IsTrue("KOOL_TTY_DISABLE") {
+	if d.Flags.DisableTty {
+		d.Warning("Warning: --disable-tty flag is obsolete")
```
fabriciojs, line 55: I guess we should not place the warning because some usage relies on something like `kool exec foo bar > xxx.sql` and having the warning will make that break... 

we could if d.Warning wrote to `Stderr` instead of straight `Stdout`... I guess removing the warning is safer - as we expect the automatic behaviour to be best than people asserting themselves whether or not to use TTY flags.

cmd/shell/terminal_test.go
fabriciojs: @danielsuguimoto OK! that is not worrying of this test... BUT, since we have `kool` being built and running on Windows, we should then add a `cmd/shell/terminal_windows.go` file with some variation for the implementation, without `creack/pty`... is there a different way we can tell this under Windows? Or maybe we can settle for having a environment variable only for Windows need.

fabriciojs (changes requested):
Looking great! We are getting there 🦾

cmd/shell/terminal_test.go
danielsuguimoto: @fabriciojs actually, this package is used on terminal testing, only. It only mocks a pseudo TTY for testing purposes.

fabriciojs: AH sorry, that is correct. `moby/terminal` will be one handling TTY assertion under Windows, we are just not checking that ourselves. That sounds good!

fabriciojs (approved):
👏🏻 👍🏻 

This will be so great for improving UX and actually simplifying using `kool` instead of straight Docker/Docker Compose!

PR #157, opened 2020-10-09
chore(editorconfig): adding config to use space as tab on yml files

danielsuguimoto:


fabriciojs (approved):
THANK YOU! this was annoying every time we were editing any YML 👍🏻

PR #162, opened 2020-10-14
feat(run): added available scripts to run command

### LINKED ISSUE #111: Proposal: List of available kool scripts
It would be good to have the possibility of showing all available kool scripts. It can be done through shell completion, a new command to list scripts or a flag `--list` on the `kool run`.

danielsuguimoto:
Closes #111

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -6,6 +6,7 @@ import (
 	"kool-dev/kool/cmd/parser"
 	"kool-dev/kool/environment"
 	"path"
+	"strings"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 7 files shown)


cmd/run.go
```diff
@@ -100,5 +96,39 @@ func NewRunCommand(run *KoolRun) (runCmd *cobra.Command) {
 
 	// after a non-flag arg, stop parsing flags
 	runCmd.Flags().SetInterspersed(false)
+
+	// look for kool.yml on current working directory
+	_ = run.parser.AddLookupPath(run.envStorage.Get("PWD"))
+	// look for kool.yml on kool folder within user home directory
+	_ = run.parser.AddLookupPath(path.Join(run.envStorage.Get("HOME"), "kool"))
+
+	if usageTempl, err := getRunUsageTemplate(run, runCmd); err == nil {
```
fabriciojs, line 105: at first I am a bit concerned of the overhead we add to the `init()` doing all the parse in order to bootstrap the commands only... but I guess UX wise it is worth. Nice :+1:

fabriciojs: Now, about ignoring the error... maybe in case we have `KOOL_VERBOSE=1` we should then print this omitted error anyway?

fabriciojs (changes requested):
Awesome!

cmd/run.go
fabriciojs: @danielsuguimoto we can look into it/evaluate in a next iteration!

fabriciojs (approved):

PR #164, opened 2020-10-15
refactor(run): refactoring run scripts

danielsuguimoto:


cmd/run.go
```diff
@@ -97,45 +104,45 @@ func NewRunCommand(run *KoolRun) (runCmd *cobra.Command) {
...
+			if run.envStorage.IsTrue("KOOL_VERBOSE") {
+				run.Println("$ got an error trying to add available scripts to command usage template; error:", err.Error())
+			}
+			return
+		}
 
-	for _, script := range scripts {
-		sb.WriteString("  ")
-		sb.WriteString(script)
+		sb.WriteString(originalUsageText)
```
fabriciojs, line 135: do we really need to have the `originalUseageText` coming from `runCmd.UsageString()` in the init() time, instead of calling it only here within this closure? So we would only generate it when actually needed... unless setting the function would generate an infinite loop, in which case it must be like it is right now.

fabriciojs (commented):
@danielsuguimoto just one question!

cmd/run.go
danielsuguimoto: @fabriciojs yep, I tried calling `UsageString()` inside the closure and it caused an infinite loop, indeed. And unfortunately, the functions to process the default template are not exported. We could replicate them on our code, but any changes on cobra code could break it.

fabriciojs: OK! we can keep as is!

fabriciojs (approved):

Issue #166, opened 2020-10-16
Unexpected --user on kool exec breaks with permission error


fabriciojs:
`kool exec` in a project with official `php:7.4.2-fpm-alpine` breaks. That is because we enter the image with a non-existing user, then any write will fail with permission issues:

```
➜  kool git:(master) docker run --rm -d php:7.4.2-fpm-alpine                                                               
64c6f469e24728d1ddbef962a6cb3c6b1bd06bc49bb552fd32376db06a2d992d
➜  kool git:(master) # works     
➜  kool git:(master) docker exec 64c6f469e24728d1ddbef962a6cb3c6b1bd06bc49bb552fd32376db06a2d992d touch /tmp/xxx
➜  kool git:(master) # do not work
➜  kool git:(master) docker exec --user 501 64c6f469e24728d1ddbef962a6cb3c6b1bd06bc49bb552fd32376db06a2d992d touch /tmp/xxx
touch: /tmp/xxx: Permission denied
```

Something created with root or any other user, when trying to access it with a non-existing user, will fail.

Currently we will always define `KOOL_ASUSER` (with current user ID), and then inject it:

https://github.com/kool-dev/kool/blob/master/cmd/exec.go#L55

We need to figure out the better UX to accommodate images that support this and images that do not.

fabriciojs:
One idea is:

- We never add the `--user` on kool exec. We add some support on the kool.yml presets file to come from there the definition. This is still not 100% because people may use the default presets with custom images, which would still cause issues.

danielsuguimoto:
@fabriciojs can't we do something that might make our own images unique somehow? Then we could detect it as we do on `kool docker` command?

fabriciojs:
@danielsuguimoto that is one way we can explore, but has its flaws... 

`kool docker` like `kool exec` have issues with options ordering when proxying to docker/docker-compose... For instance I currently have use cases where I need more flags to be send to `docker run` so I do like:

`kool docker -- --network xxx image:tag command`

So I am able to define the `network` even though we did not create such option. But in this case we no longer no what of the arguments in the list is the action target image.

Note: `--` as a parameters tells the command to stop parsing any flags, start treating them all as arguments.

We need another way to "check" if we need/want or not to set the user. UX-wise. That  is the big question indeed.

fbrissi:
It's the same problem I had in a other project with Symfony, in my case it was because of Windows.

[...]

PR #168, opened 2020-10-16
feat(root): Added verbose flag to all commands

### LINKED ISSUE #167: Add --verbose flag to root command
**Describe the feature**

Increased verbosity should also be a achieve through a most usual manner like `kool --verbose ...`, instead than relying on environment varialbe `KOOL_VERBOSE=1` only.

We should keep accepting `KOOL_VERBOSE`, but in case of verbosity enabled (used), we output that is deprecated. We can remove this support on 2.0.

**Use cases**

Easier documentation `kool --help` would already print it; usual way of getting more verbose output in general CLI usage.

danielsuguimoto:
Closes #167

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -29,7 +29,7 @@ Complete documentation is available at https://kool.dev/docs`,
 		},
 	}
 
-	cmd.PersistentFlags().Bool("verbose", false, "execute commands with verbosity")
+	cmd.PersistentFlags().Bool("verbose", false, "increases output verbosity")
 	return
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -3,7 +3,6 @@ package cmd
 import (
 	"github.com/spf13/cobra"
 	"kool-dev/kool/environment"
-	"os"
 )
 
 // CobraRunFN Cobra command run function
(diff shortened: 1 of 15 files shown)


cmd/root.go
```diff
@@ -2,21 +2,41 @@ package cmd
...
+			}
+
+			if asuser := cmf.Flags().Lookup("asuser"); asuser != nil && asuser.Value.String() != "0" {
+				envStorage.Set("KOOL_ASUSER", asuser.Value.String())
+			}
+		},
+	}
+
+	cmd.PersistentFlags().Bool("verbose", false, "execute commands with verbosity")
+	cmd.PersistentFlags().Int("asuser", os.Getuid(), "execute commands as user")
```
fabriciojs, line 38: The `--verbose` flag I guess is worth having it persisted, but the `asuser` do not make sense for most of the commands we have, and also it is under review on #166. So I would say we should not touch the `asuser` at this moment, removing it from this change set that otherwise looks good.

fabriciojs (changes requested):
I guess we should not handle asuser like this at this  moment, the rest looks positively good!

cmd/root.go
```diff
@@ -2,21 +2,35 @@ package cmd
...
+		Version:           version,
+		DisableAutoGenTag: true,
+		PersistentPreRun: func(cmf *cobra.Command, args []string) {
+			if verbose := cmf.Flags().Lookup("verbose"); verbose != nil && verbose.Value.String() == "true" {
+				envStorage.Set("KOOL_VERBOSE", verbose.Value.String())
+			}
+		},
+	}
+
+	cmd.PersistentFlags().Bool("verbose", false, "execute commands with verbosity")
```
fabriciojs, line 32: One thing, for the sake of language, we are *increasing* verbosity, so maybe it is better something like - `increses output verbosity` for description.

PR #170, opened 2020-10-19
feat(task): added new long task runner

### LINKED ISSUE #169: Improve commands long tasks output
**Describe the feature**

Improve long tasks (as `kool self-update`) output

Would this be BC? No

**Use cases**

Show the user that a long task is running.

**Examples**

When you run `kool self-update`, nothing is printed, unless we have a success or a failure

danielsuguimoto:
Closes #169


cmd/task/task.go
```diff
@@ -0,0 +1,50 @@
...
+	r.out.Printf("%s ... ", message)
+	chError := make(chan error)
+
+	go func() {
+		chError <- closure()
+		wg.Done()
+	}()
+
+	err = <-chError
+	wg.Wait()
```
fabriciojs, line 39: there is no need to have the WaitGroup at this flow, since the channel reading already gives us the blocking we want.

cmd/task/task.go
```diff
@@ -0,0 +1,50 @@
...
+
+	r.out.Printf("%s ... ", message)
+	chError := make(chan error)
+
+	go func() {
+		chError <- closure()
+		wg.Done()
+	}()
+
+	err = <-chError
```
fabriciojs, line 38: A nice feature we may add soon is to have optional timeout setting to the task, so we can at this point not only wait for the error return on chError, but also a timer:

```go
select {
   case err = <-chError:
       ...
   case <-time.After(r.timeout):
       // timeout!
}
```

cmd/task/task.go
```diff
@@ -0,0 +1,50 @@
...
+func NewRunner() Runner {
+	return &DefaultRunner{shell.NewOutputWriter()}
+}
+
+// Run runs task
+func (r *DefaultRunner) Run(message string, closure func() error) (err error) {
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	r.out.Printf("%s ... ", message)
```
fabriciojs, line 30: this is a nice feature to start with, but we need to advise it only works for no-output tasks. If the closure() has some kind of output, this would get messy.

I would like to try exploring something like:

- We have a pipe for capturing the task actual output.
- While executing a task, we have something like:

```
Running "task"...
-------------------------------
(original task output... keeps writing here)
-------------------------------
... (animated char, something like an ASCII spinner with `-\|/-` replacing sequence, or something for that matter.
```


So the last output line is ALWAYS our "progress/running" spinner, while above it we keep displaying any eventual output the task could have.

Not sure how to architecture that, we may discuss further, maybe for a future iteration?

cmd/self-update.go
```diff
@@ -27,17 +29,22 @@ func init() {
...
 	}
 }
 
 // Execute runs the self-update logic with incoming arguments.
 func (s *KoolSelfUpdate) Execute(args []string) (err error) {
-	var latestVersion semver.Version
+	var currentVersion, latestVersion semver.Version
 
-	currentVersion := s.updater.GetCurrentVersion()
+	err = s.taskRunner.Run("Updating kool version", func() (taskError error) {
```
fabriciojs, line 41: I understand the practice this gives  us, but I am a bit worried there is a hidden race condition where we don't know for sure if the closure was ever executed, but we do conditionally behave below on something that depends on it. Maybe I would suggest getting all the behaviour below and bring it to within this closure.

fabriciojs (changes requested):
Awesome work on more test coverage!

Couple of comments, which I believe we should address 2 and other 2 we may leave for the next iteration on this!

cmd/task/task.go
fabriciojs: For an eventual spinner: https://github.com/briandowns/spinner :D

cmd/self-update.go
```diff
@@ -27,26 +30,40 @@ func init() {
...
 		*newDefaultKoolService(),
+		task.NewRunner(),
 		&updater.DefaultUpdater{RootCommand: rootCmd},
 	}
 }
 
 // Execute runs the self-update logic with incoming arguments.
 func (s *KoolSelfUpdate) Execute(args []string) (err error) {
-	var latestVersion semver.Version
+	newVersion, err := s.taskRunner.Run("Updating kool version", func() (taskResult interface{}, taskError error) {
```
fabriciojs, line 40: Having the return `taskResult interface{}` is a bit odd, I would rather try to make this in a way the task.Runner would wrap cmd.KoolService and encapsulate the cmd.KooService.Execute call with proper outout/error handling.

cmd/kool_task.go
```diff
@@ -0,0 +1,72 @@
...
+	bufBytes, _ := ioutil.ReadAll(buf)
+	output = strings.TrimSpace(string(bufBytes))
+
+	if err != nil {
+		t.Println(color.New(color.Red).Sprint("error"))
+	} else {
+		t.Println(color.New(color.Green).Sprint("done"))
+	}
+
+	fmt.Println(output)
```
fabriciojs, line 54: guess this one was left by mistake?

cmd/kool_task.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	return
+}
+
+func (t *DefaultKoolTask) execService(args []string) <-chan error {
+	chError := make(chan error)
+
+	go func() {
+		defer close(chError)
```
fabriciojs, line 67: are we guaranteed to have the written value being read, even if the deferred close happens before the read at line 42? I mean, race condition wise?

fabriciojs (changes requested):
Good job!

Two minor comments so we can merge this!

cmd/kool_task.go
fabriciojs: So yeah it is secure!

image

cmd/kool_task.go
danielsuguimoto: Big oops!

fabriciojs (approved):

Issue #11, opened 2020-08-04
Proposal: kool preset [preset] --wizard | Tailor the presets


fabriciojs:
Would be great to have the `kool init` flexible enough for people to use different database services, or cache services, or even remove those if they don't wanna use them.

---

Nice tool for CLI input: https://github.com/AlecAivazis/survey

---

### TODO

- [x] Given the chosen preset we need to know the options that are adjustable interactively. Figure out a way to have this.

- [x] Provided the chosen options, we then need to build up the final configuration based on the preset templates, that will have to be split up in order to ease up this process.

fabriciojs:
Use cases:

```
$ kool init
     > What preset you wanna use?
         > Laravel
            Adonis
            NextJS
            NuxtJS
      > What database service you wanna use?
          > Mysql
              MariaDB
              Postgress
              None
      > What cache service you wanna use?
          > Redis
              Memcached
              None
```

danielsuguimoto:
another package to consider using: https://github.com/manifoldco/promptui

fabriciojs:
@danielsuguimoto nice!

We also should think about what options we are willing to give initially...

It is important we have them tested and working, so, fewer stable options is certainly better than a lot of options with hiccups in the way.

Starting to think for Laravel for example:

**Databases**
- Mysql (5.7, 8.0)
- Postgress (version?)
- MariaDB (version?)
- None - we do not put in a `database` service

**Application**
- PHP (7.4, 7.3, 7.2)

fabriciojs:
@danielsuguimoto right now I see we have "scrolling" in the list.

![image](https://user-images.githubusercontent.com/1116377/96151587-f4a7cc00-0ee1-11eb-8456-85c241d79a96.png)


The filtering feature is really nice. I guess we should just think about an UX where we avoid more than 5-6 options at a time, so no scrolling happens. (i.e asking something before, like, `project language: PHP, JS, Go`.

danielsuguimoto:
@fabriciojs do we change presets folders structure? Or do you have any idea in mind to relate presets to languages, for example?

fabriciojs:
@danielsuguimoto now that we have https://github.com/kool-dev/kool/pull/172 merged, this would be a breaking change... I think we make it to `2.0.0` or we could add back a command `init` that prompts a deprecated message and proxies to kool preset. What do you think?

danielsuguimoto:
[...]

PR #172, opened 2020-10-20
refactor(preset): Refactoring init command (now preset command)

danielsuguimoto:
Relates to #11 

* Renamed `kool init` to `kool preset`
* Added language question before asking the preset
* Presets parsing now is made with a go script, not in bash anymore


dbpolito:
❤️

cmd/presets/presets.go
```diff
@@ -1,10 +1,14 @@
 package presets
 
 // auto generated file
+
 // GetAll get all presets
 func GetAll() map[string]map[string]string {
 	var presets = make(map[string]map[string]string)
 	presets["adonis"] = map[string]string{
+		".dockerignore": `/node_modules
```
fabriciojs, line 9: hmm, don't we actually need the node_modules within the built image? Not sure where this line came from.

kool.yml
```diff
@@ -22,4 +22,7 @@ scripts:
   # Generate documentation for kool commands
   make-docs:
     - rm -f docs/4-Commands/*.md
-    - kool docker golang:1.15.0 go run docs.go
+    - kool docker --volume=kool_gopath:/go golang:1.15.0 go run docs.go
+  parse-presets:
```
fabriciojs, line 26: did this simply override the `parse-presets` above? maybe we should remove that one hehe

fabriciojs: image

fabriciojs (changes requested):
Awesome!

cmd/preset.go
```diff
@@ -27,29 +27,33 @@ var ErrPresetFilesAlreadyExists = errors.New("some preset files already exist")
...
 		shell.NewPromptSelect(),
 	}
 }
 
-// Execute runs the init logic with incoming arguments.
-func (i *KoolInit) Execute(args []string) (err error) {
-	var fileError, preset string
+// Execute runs the preset logic with incoming arguments.
+func (i *KoolPreset) Execute(args []string) (err error) {
+	var fileError, preset, language string
```
fabriciojs, line 49: right now this command pretty much NEEDS an TTY to execute. I guess would be nice to add a check for it, and error out with a proper message in case it is running under not-TTY environment.

cmd/preset.go
```diff
@@ -84,30 +88,30 @@ func (i *KoolInit) Execute(args []string) (err error) {
...
-// NewInitCommand initializes new kool init command
-func NewInitCommand(init *KoolInit) (initCmd *cobra.Command) {
-	initCmd = &cobra.Command{
-		Use:   "init [PRESET]",
-		Short: "Initialize kool preset in the current working directory",
+// NewPresetCommand initializes new kool preset command
+func NewPresetCommand(preset *KoolPreset) (presetCmd *cobra.Command) {
+	presetCmd = &cobra.Command{
+		Use:   "preset [PRESET]",
+		Short: "Initialize kool preset in the current working directory. If none is specified, a wizard will be prompt to choose one",
```
fabriciojs, line 95: `If no preset argument is specified you will be prompted to pick among the existing options.`

cmd/preset.go
```diff
@@ -84,30 +88,30 @@ func (i *KoolInit) Execute(args []string) (err error) {
 	return
 }
 
-// NewInitCommand initializes new kool init command
```
fabriciojs, line 87: Above, at line 82:

image


I may have overlooked how you dealt with this, but won't we start creating a `preset_language` file as well now? I guess we DO NOT want that, so we need to have some kind of check to avoid doing it?

cmd/presets/presets.go
danielsuguimoto: @fabriciojs hummm...`.dockerignore` file was always inside the preset folder. But bash scripts can't read files which names start with `.`. Now I created a go script to do it, and go can read this kind of file. To remove it from here, we should remove `.dockerignore from `adonis` folder

kool.yml
danielsuguimoto: ooops

cmd/presets/presets.go
danielsuguimoto: @fabriciojs keep them? Remove `node_modules` from all `.dockerignore`? I guess it was a bug from `parse_presets.sh` that wasn't parsing these files.

dbpolito: we want to keep it...

we. want dockerignore so on COPY it won't inject your node_modules to the container build and build from scratch... i got a bug because node_modules had different symlink and broke everything...

fabriciojs: @danielsuguimoto oh I see! so never mind my comment, as @dbpolito pointed this indeed is a fix! 👏🏻

cmd/preset.go
danielsuguimoto: @fabriciojs it only needs TTY when no args are passed. Could we validate only for this situation?

fabriciojs: Yes! that sounds correct, only when needed to ask for the preset!

kool.yml
```diff
@@ -22,4 +19,8 @@ scripts:
   # Generate documentation for kool commands
   make-docs:
     - rm -f docs/4-Commands/*.md
-    - kool docker golang:1.15.0 go run docs.go
+    - kool docker --volume=kool_gopath:/go golang:1.15.0 go run docs.go
+  # Parsing the preset files onto Go code in a shell automated fashion.
```
fabriciojs, line 23:
```suggestion
  # Parsing the preset files onto Go code in an automated fashion.
```

No longer `shell` 😄

fabriciojs (approved):

Issue #42, opened 2020-09-02
Proposal: Windows support


fabriciojs:
Right now the best way to use Kool on Windows is actually through WSL2 - which is pretty much a Linux environment so things should work mostly out-of-the-box.

Now running under Windows is a big question mark due to environment differences - executing docker, docker-compose, flags parsing, etc.

We have people testing that in an experimental fashion at this moment.

fbrissi:
I stiil testing this, but it seems to be very promising for now.

![image](https://user-images.githubusercontent.com/339706/92105232-eb8eef80-edb8-11ea-8cc7-4bd872b92b57.png)
![image](https://user-images.githubusercontent.com/339706/92105259-f6e21b00-edb8-11ea-9859-b1f7bd191aeb.png)
![image](https://user-images.githubusercontent.com/339706/92105375-23963280-edb9-11ea-8639-4886664e962c.png)
![image](https://user-images.githubusercontent.com/339706/92105330-0feacc00-edb9-11ea-94ae-4781a3d82d25.png)

fabriciojs:
This has been added on latest release 1.0.16.

But we have an issue - Windows tells the installer is a security risk. In order to solve that we need to sign the program.

@fbrissi pointer that out: https://docs.microsoft.com/pt-br/windows/msix/package/sign-app-package-using-signtool

fabriciojs:
Windows store - can we delivery through there?

wesleyhf:
I think windows store is a good way to go, it will keep everything simple, just go to store and click install!

https://docs.microsoft.com/en-us/windows/uwp/publish/

fabriciojs:
@cesramires - your notes about WSL are great - maybe you could provide a PR to add this somewhere in the docs/  markdown files? Installation section, maybe a new page linked, with WSL specifics.

fabriciojs:
About Windows store, I am moving this to be analysed at https://github.com/kool-dev/kool/issues/89.

PR #175, opened 2020-10-22
Fix UID Windows

fbrissi:
Correction of generating uid for Windows and installing does not add kool on Windows startup.

- As in Windows we don't have uid, use sid and get the last number, example: S-1-5-21-3623811015-3361044348-30300820-1013
 So, in this case, the uid used is 1013, this solve the problem to use root,  but it still has problems as in other cases that we saw in linux in docker images
 - Fix the installer was putting the kool to boot with windows start
 - Add arguments to build_artifacts, sample: `./build_artifacts.sh --version 1.6.0`

### DIFF
--- a/environment/uid_windows.go
+++ b/environment/uid_windows.go
@@ -1,0 +1,25 @@
+package environment
+
+import (
+	"os/user"
+	"regexp"
+)
+
+var sidExp = regexp.MustCompile(`([\d\D]+-)(?P<uid>\d+)`)
+
+func initUid(envStorage EnvStorage) {
+	// under native windows defaults to using
+	// root inside containers for kool managed images
+	envStorage.Set("UID", uid())
+}
+
+func uid() string {
+	current, _ := user.Current()
+	match := sidExp.FindStringSubmatch(current.Uid)
+
+	results := map[string]string{}
+	for i, name := range match {
+		results[sidExp.SubexpNames()[i]] = name
+	}
+	return results["uid"]
+}
--- a/environment/uid_windows.go
+++ b/environment/uid_windows.go
@@ -5,7 +5,7 @@ import (
 	"regexp"
 )
 
-var sidExp = regexp.MustCompile(`([\d\D]+-)(?P<uid>\d+)`)
+var sidExp = regexp.MustCompile(`.*-(?P<uid>\d+)`)
 
 func initUid(envStorage EnvStorage) {
 	// under native windows defaults to using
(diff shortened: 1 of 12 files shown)


danielsuguimoto (commented):
Can't we merge `asuser` and `uid` files? They are already sharing the `uid()` function, so keeping both in a single file could be more readable, IMO. But let's check the other's opinions

environment/uid_windows.go
```diff
@@ -0,0 +1,25 @@
+package environment
+
+import (
+	"os/user"
+	"regexp"
+)
+
+var sidExp = regexp.MustCompile(`([\d\D]+-)(?P<uid>\d+)`)
```
danielsuguimoto, line 8: If it's being used just inside `uid()`, there is no need to keep it in package scope level. You can declare it inside `uid()`

danielsuguimoto (changes requested):

environment/uid_windows.go
fbrissi: [MustCompile](https://golang.org/pkg/regexp/#MustCompile) is similar [Compile](https://golang.org/pkg/regexp/#Compile), this compile and parses a regular expression

I think it’s not a good idea to compile every time it’s called

danielsuguimoto: hum...yeah...it makes sense...never mind...

environment/uid_windows.go
```diff
@@ -0,0 +1,25 @@
...
+
+func uid() string {
+	current, _ := user.Current()
+	match := sidExp.FindStringSubmatch(current.Uid)
+
+	results := map[string]string{}
+	for i, name := range match {
+		results[sidExp.SubexpNames()[i]] = name
+	}
+	return results["uid"]
```
fabriciojs, line 24: what is the change we run this in a different version of windows, and for some reason we do not get a proper "uid" match from FindStringSubmatch above?

I guess we should here check IF the "uid" key exists, and then return its value; other wise I guess we should not break but instead we can default to the old behaviour?

fabriciojs (changes requested):
Awesome! Thanks for improving the Windows experience!

Just one note about resilience to avoid errors on unexpected environments.

fabriciojs:
This is related to #42

environment/uid_windows.go
fbrissi: I did it.
I tried to do a unit test to get this error, but I couldn't, I commented where the test failed. 
I wanted to simulate a case of divine work. 😄

fabriciojs (approved):

fabriciojs:
@danielsuguimoto you can review and then merge!

danielsuguimoto (approved):
LGTM

PR #177, opened 2020-10-22
feat(init): deprecating init command

danielsuguimoto:
Relates to #11 

This pull request brings back the init command, but deprecated and working as a proxy to the preset command, in case we need to release a new tag before `2.0.0`

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -1,0 +1,12 @@
+package cmd
+
+import "github.com/gookit/color"
+
+func init() {
+	initCmd := NewPresetCommand(NewKoolPreset())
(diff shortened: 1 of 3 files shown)


fabriciojs (approved):

fabriciojs:
@danielsuguimoto thanks! this is good so we can have 1.6.0 prior, and wait for 2.0 with other bigger changes - we should use major versions to really do most of the breaking changes we can at once :D

fabriciojs:
@danielsuguimoto now that this one is on `master`, I will create `master` -> `2.0`; after that, we should create a new PR removing the init proxy once again, so 2.0 do not keep it.

PR #181, opened 2020-10-23
changed preset to init

sineld:


fabriciojs:
@sineld thanks for the PR!

I am imagining what prompted this - you were following the documentation on Github, using latest kool version `1.5.0`, and then realised it mentioned a `preset` command that didn't exist, and realised it was meant to be `init` - is that what happened?

Just to put a context on this - for semantic reasons we are changing `init` to `preset`, that is already merged onto `master` but not yet released in a new version. We will still keep `kool init` but as deprecated, so it will only be removed when 2.0 comes out. This work is part of #11 

So as soon as we release `1.6.0` we will have `kool preset` (still will have `kool init`, but noted as deprecated).

@danielsuguimoto - let's do a release as soon as possible!

fabriciojs:
@dbpolito - the GH action for updating kool.dev/docs is happening whenever we go to `master`. Can we change it to happen only for tags creation?

image

sineld:
Cool, I just thought iw was written wrong.
Well done.
Thanks.

danielsuguimoto:
Version `1.6.0` has just been released. Now docs and kool are in sync. I'm closing this pull request, but thanks for contributing anyway, @sineld!

dbpolito:
FYI: i fixed the GH action to avoid this happening again https://github.com/kool-dev/kool/commit/b487f7976458974cc3f6ec88c3259c946a0d2785

Thanks again for pointing it for us :+1:

PR #187, opened 2020-10-27
feat(task): adding spinner to kool task

### LINKED ISSUE #180: Long running task progress/real time output
this is a nice feature to start with, but we need to advise it only works for no-output tasks. If the closure() has some kind of output, this would get messy.

I would like to try exploring something like:

- We have a pipe for capturing the task actual output.
- While executing a task, we have something like:

```
Running "task"...
-------------------------------
(original task output... keeps writing here)
-------------------------------
... (animated char, something like an ASCII spinner with `-\|/-` replacing sequence, or something for that matter.
```

So the last output line is ALWAYS our "progress/running" spinner, while above it we keep displaying any eventual output the task could have.

Not sure how to architecture that, we may discuss further, maybe for a future iteration?

_Originally posted by @fabriciojs in https://github.com/kool-dev/kool/pull/170#discussion_r508751614_

danielsuguimoto:
Related to #180

### DIFF
--- a/cmd/kool_task.go
+++ b/cmd/kool_task.go
@@ -2,9 +2,10 @@ package cmd
 
 import (
 	"bufio"
-	"bytes"
 	"fmt"
-	"os"
+	"io"
+	"kool-dev/kool/cmd/shell"
+	"strings"
 	"time"
 
 	"github.com/gookit/color"
@@ -20,11 +21,12 @@ type KoolTask interface {
 type DefaultKoolTask struct {
 	KoolService
 	message string
+	taskOut shell.OutputWriter
 }
 
 // NewKoolTask creates a new kool task
 func NewKoolTask(message string, service KoolService) *DefaultKoolTask {
-	return &DefaultKoolTask{service, message}
+	return &DefaultKoolTask{service, message, shell.NewOutputWriter()}
 }
 
 // Run runs task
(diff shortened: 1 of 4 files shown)


cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+	scanner.Split(bufio.ScanLines)
+
+	go func() {
+		defer func() {
+			close(lines)
+			close(doneScanning)
+		}()
+
+		for range time.Tick(100 * time.Millisecond) {
+			if !scanner.Scan() {
```
fabriciojs, line 95: let's check how it goes if we do `for scanner.Scan() ` instead of using the time.Tick above for throttling.

fabriciojs: we need to be aware of possible CPU high usage

cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+	spinChars := []byte{'-', '/', '|', '\\'}
+	spinPos := 0
 
 	go func() {
-		defer close(chError)
-		chError <- t.Execute(args)
+		defer close(donePrinting)
+
+	OutputPrint:
+		for range time.Tick(100 * time.Millisecond) {
```
fabriciojs, line 117: `for {`

infinite loop

cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+			spinPos = (spinPos + 1) % 4
+			currentSpin := spinChars[spinPos : spinPos+1]
+
+			select {
+			case <-doneScanning:
```
fabriciojs, line 122: as said before, we may lose this for checking the end

cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+				// remaining lines
+				for line := range lines {
+					if line != "" {
+						t.taskOut.Println(">", line)
+					}
+				}
+
+				t.taskOut.Printf("... %s", currentSpin)
+				break OutputPrint
+			case line := <-lines:
```
fabriciojs, line 134: here we may read two values out of the channel... `line, ok := <-lines` where `line` is the value, and `ok` is whether the channel was closed or not.

fabriciojs: image

cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+				spinPos = (spinPos + 1) % 4
+
+				t.taskOut.Printf("\r")
+
+				if line != "" {
+					t.taskOut.Println(">", line)
+				}
+
+				t.taskOut.Printf("... %s", currentSpin)
+			default:
```
fabriciojs, line 144: instead of `default:` we may do a `<-time.After(100*time.Millisecond)`  and then within it only we do the spinning.

cmd/kool_task.go
```diff
@@ -30,40 +35,119 @@ func (t *DefaultKoolTask) Run(args []string) (err error) {
...
+		defer close(err)
+		err <- t.Execute(args)
+	}()
+
+	return err
+}
+
+func startServiceOutputScanner(reader io.Reader, lines chan string) <-chan bool {
+	doneScanning := make(chan bool)
+	scanner := bufio.NewScanner(reader)
```
fabriciojs, line 85: maybe instead of a bufio.Scanner using a bufio.Reader with ReadString is best (it seems to be blocking and then we can wait on it to feed the lines)

cmd/kool_task.go
fabriciojs: https://golang.org/pkg/bufio/#Reader.ReadString could be an option too

fabriciojs (changes requested):
Some comments for assessments

fabriciojs (approved):
Looking good! but let's do some testing before release - going to add to the issue.

PR #196, opened 2020-11-13
feat(preset): Add wizard to customiza docker-compose.yml

danielsuguimoto:
Relates to #11


templates/database/prostgresql.yml
```diff
@@ -0,0 +1,12 @@
+image: postgres
```
dbpolito, line 1: maybe we should add the version here... `13`

dbpolito: also i see they have an alpine version, maybe we should use it too

cmd/presets/presets.go
```diff
@@ -508,8 +511,11 @@ networks:
 		".dockerignore": `/node_modules
 /vendor
 `,
-		"preset_language": "php",
-		"preset_create":   "kool docker kooldev/php:7.4 composer create-project --prefer-dist symfony/website-skeleton",
+		"preset_language":         "php",
+		"preset_create":           "kool docker kooldev/php:7.4 composer create-project --prefer-dist symfony/website-skeleton",
+		"preset_ask_services":     "database,cache",
+		"preset_database_options": "MySQL 8.0,MySQL 5.7,ProstgreSQL 13.0,none",
```
fabriciojs, line 517: Also for a next iteration, I would love to change this somehow into an YML config structure, for better structure.

cmd/presets/parser.go
```diff
@@ -91,56 +111,86 @@ func (p *DefaultParser) LookUpFiles(preset string) (foundFiles []string) {
...
+	sort.Strings(keys)
+
+	return
+}
+
+// GetPresetKeyContent get preset key value
+func (p *DefaultParser) GetPresetKeyContent(preset string, key string) (value string) {
+	presetData := p.Presets[preset]
+
+	for dataKey, dataContent := range presetData {
```
fabriciojs, line 173: presetData is a map right? better to do `dataContent , found = presetData[dataKey]` then.

cmd/preset.go
```diff
@@ -124,3 +194,10 @@ func NewPresetCommand(preset *KoolPreset) (presetCmd *cobra.Command) {
 	presetCmd.Flags().BoolVarP(&preset.Flags.Override, "override", "", false, "Force replace local existing files with the preset files")
 	return
 }
+
+func formatTemplateKey(key string) (formattedKey string) {
```
fabriciojs, line 198: took me a while to track the code path till here... I can see how it works, but we do need to improve/refactor. This for example should not be necessary if we have better structured  presets data to start with.

fabriciojs: This is too indirect/not-intuitive.

fabriciojs (approved):
@danielsuguimoto this is a super feature!

I do like the overall structure. As you can see in my notes I believe we can greatly improve 1. the structure we store the options; 2. the code path for checking all possibilities (extract into other services).

For now I think we are good to start testing it as is (let's hold off the release until we do at least one iteration in the mentioned improvements).

Issue #17, opened 2020-08-07
Proposal: kool create


fabriciojs:
Create a new project for the presets we already handle.

For example, `kool create laravel my-laravel-app` would create a new project laravel from scratch using composer project-create in a `my-laravel-app` folder in the current working dir.

danielsuguimoto:
I've added a metafile with some info about the preset on this pull request https://github.com/kool-dev/kool/pull/172
With this metafile, we will be able to set a create-project command as well.

fabriciojs:
Awesome! @tiagoguirado or @nkabz  do you wanna tackle this one? 😃 👀

leandroncbrito:
I'll tackle this one

dbpolito:
Thanks @leandroncbrito for your contribution, looking great.

PR #198, opened 2020-11-16
kool create command

leandroncbrito:
https://github.com/kool-dev/kool/issues/17

New command "create"

**Parameters:** 
1. preset/framework name
2. project/folder name

Ex: kool create laravel my-app

**New issue improvement:** 
Instead of entering the preset name, we might show the preset menu and let the user choose.

### DIFF
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"fmt"
 	"kool-dev/kool/cmd/presets"
+	"os"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -13,6 +14,7 @@ type KoolCreate struct {
 	DefaultKoolService
 	parser presets.Parser
 	KoolDocker
+	KoolPreset
 }
 
 func init() {
@@ -30,12 +32,14 @@ func NewKoolCreate() *KoolCreate {
 		*newDefaultKoolService(),
 		&presets.DefaultParser{Presets: presets.GetAll()},
 		*NewKoolDocker(),
+		*NewKoolPreset(),
 	}
 }
 
 // Execute runs the exec logic with incoming arguments.
 func (c *KoolCreate) Execute(originalArgs []string) (err error) {
 	preset := originalArgs[0]
+	dir := originalArgs[1]
 
 	if !c.parser.Exists(preset) {
 		err = fmt.Errorf("Unknown preset %s", preset)
@@ -50,7 +54,23 @@ func (c *KoolCreate) Execute(originalArgs []string) (err error) {
 
 	args := append(strings.Fields(createCmd), originalArgs[1:]...)
 
-	c.KoolDocker.Execute(args)
+	err = c.KoolDocker.Execute(args)
+
+	if err != nil {
+		return
+	}
+
+	err = os.Chdir(dir)
+
+	if err != nil {
+		return
+	}
+
+	err = c.KoolPreset.Execute([]string{preset})
+
+	if err != nil {
+		return
+	}
 
 	return
 }
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"kool-dev/kool/cmd/builder"
 	"kool-dev/kool/cmd/presets"
 	"strings"
 
@@ -12,8 +11,8 @@ import (
 // KoolCreate holds handlers and functions to implement the preset command logic
 type KoolCreate struct {
 	DefaultKoolService
-	parser     presets.Parser
-	koolDocker builder.Command
+	parser presets.Parser
+	KoolDocker
 }
 
 func init() {
@@ -30,17 +29,13 @@ func NewKoolCreate() *KoolCreate {
 	return &KoolCreate{
 		*newDefaultKoolService(),
 		&presets.DefaultParser{Presets: presets.GetAll()},
-		builder.NewCommand("kool", "docker"),
+		*NewKoolDocker(),
 	}
 }
 
 // Execute runs the exec logic with incoming arguments.
-func (c *KoolCreate) Execute(args []string) (err error) {
-	if !c.IsTerminal() {
-		c.koolDocker.AppendArgs("-T")
-	}
-
-	preset := args[0]
+func (c *KoolCreate) Execute(originalArgs []string) (err error) {
+	preset := originalArgs[0]
 
 	if !c.parser.Exists(preset) {
 		err = fmt.Errorf("Unknown preset %s", preset)
@@ -53,9 +48,9 @@ func (c *KoolCreate) Execute(args []string) (err error) {
 		return
 	}
 
-	c.koolDocker.AppendArgs(strings.Fields(createCmd)...)
+	args := append(strings.Fields(createCmd), originalArgs[1:]...)
 
-	err = c.koolDocker.Interactive(args[1:]...)
+	c.KoolDocker.Execute(args)
 
 	return
 }
--- a/cmd/create.go
+++ b/cmd/create.go
@@ -30,7 +30,7 @@ func init() {
 func NewKoolCreate() *KoolCreate {
 	return &KoolCreate{
 		*newDefaultKoolService(),
-		&presets.DefaultParser{Presets: presets.GetAll()},
+		&presets.DefaultParser{},
 		*NewKoolDocker(),
 		*NewKoolPreset(),
 	}
@@ -40,6 +40,8 @@ func NewKoolCreate() *KoolCreate {
 func (c *KoolCreate) Execute(originalArgs []string) (err error) {
 	preset := originalArgs[0]
 	dir := originalArgs[1]
+
+	c.parser.LoadPresets(presets.GetAll())
 
 	if !c.parser.Exists(preset) {
 		err = fmt.Errorf("Unknown preset %s", preset)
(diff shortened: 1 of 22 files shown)


dbpolito:
@leandroncbrito after you create, it's already having the kool files there? or you still need to `init/preset` afterwards?

I think we can safely create and init.

cmd/create.go
```diff
@@ -0,0 +1,73 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+// KoolCreate holds handlers and functions to implement the preset command logic
+type KoolCreate struct {
+	DefaultKoolService
+	parser     presets.Parser
+	koolDocker builder.Command
```
danielsuguimoto, line 16: Instead of a command builder calling `kool docker`, you can use the `KoolDocker` service.

cmd/create.go
```diff
@@ -0,0 +1,73 @@
...
+		&presets.DefaultParser{Presets: presets.GetAll()},
+		builder.NewCommand("kool", "docker"),
+	}
+}
+
+// Execute runs the exec logic with incoming arguments.
+func (c *KoolCreate) Execute(args []string) (err error) {
+	if !c.IsTerminal() {
+		c.koolDocker.AppendArgs("-T")
+	}
```
danielsuguimoto, line 41: `KoolDocker` already handles it if you use it instead of a command builder.

cmd/presets/fake_parser.go
```diff
@@ -19,6 +20,13 @@ func (f *FakeParser) Exists(preset string) (exists bool) {
 	return
 }
 
+// GetCreateCommand gets the command to create a new project
+func (f *FakeParser) GetCreateCommand(preset string) (cmd string, err error) {
```
danielsuguimoto, line 24: Add a test case for this new function in `fake_parser_test.go`

danielsuguimoto (changes requested):

danielsuguimoto:
> @leandroncbrito after you create, it's already having the kool files there? or you still need to `init/preset` afterwards?
> 
> I think we can safely create and init.

I like the idea too. The only question I have: should we add an option for creating the project only, without copying the preset files?

dbpolito:
Yeah but i think it makes sense creating it as default... maybe a `--no-preset` flag to disable it?

fabriciojs:
> Yeah but i think it makes sense creating it as default... maybe a `--no-preset` flag to disable it?

No need for this flag IMO. Design wise, we want people to use our environment. Even if they do not want to, having the files installed is not a big deal - no issues should arise from that.

leandroncbrito:
Yeah, I like the idea. 
I'll make the code improvements that Jun left and take a look at this change.

danielsuguimoto (commented):
Nice! I would add just one test case to the `create` command. If we add a test case and mock the preset fake parser to return true on the method `Exists` (set `MockExists` to true), then we can improve the test coverage. This way I guess all checks for this PR will pass.

danielsuguimoto (approved):
LGTM! Thx!

danielsuguimoto (changes requested):
@leandroncbrito sorry, I was almost forgetting about the docs. Please, run `kool run make-docs` to generate the docs markdown to this new command, then I think it's ready to go.

leandroncbrito:
> @leandroncbrito sorry, I was almost forgetting about the docs. Please, run `kool run make-docs` to generate the docs markdown to this new command, then I think it's ready to go.

done

docs/4-Commands/0-kool.md
```diff
@@ -16,10 +16,11 @@ Complete documentation is available at https://kool.dev/docs
 ```
 
 ### SEE ALSO
+
+* [kool create](kool-create.md)	 - Create a new project using preset
 * [kool docker](kool-docker.md)	 - Creates a new container and runs the command in it.
 * [kool exec](kool-exec.md)	 - Execute a command within a running service container
 * [kool info](kool-info.md)	 - Prints out information about kool setup (like environment variables)
-* [kool init](kool-init.md)	 - [DEPRECATED] Proxies preset command
```
danielsuguimoto, line 22: I guess the script to generate the docs has an issue with this proxied command and it's removing it by mistake. It will be removed only at version `2.0.0`, so we should keep it here, for now...

danielsuguimoto (changes requested):

docs/4-Commands/kool-init.md
```diff
@@ -1,6 +1,6 @@
 ## kool init
 
-[DEPRECATED] Initialize kool preset in the current working directory. If no preset argument is specified you will be prompted to pick among the existing options.
```
leandroncbrito, line 3: @danielsuguimoto is it the same case here? Should I revert this change?

danielsuguimoto: @leandroncbrito yep

danielsuguimoto (approved):

danielsuguimoto:
LGTM! What do you think @dbpolito @fabriciojs ?

dbpolito (approved):

cmd/presets/parser.go
```diff
@@ -29,6 +30,21 @@ func (p *DefaultParser) Exists(preset string) (exists bool) {
...
+// ErrCreateCommandtNotFoundOrEmpty error throwed when did not find the preset create command or it's empty
+var ErrCreateCommandtNotFoundOrEmpty = errors.New("create command not found or empty")
+
+// GetCreateCommand gets the command to create a new project
+func (p *DefaultParser) GetCreateCommand(preset string) (cmd string, err error) {
+	cmd = p.Presets[preset]["preset_create"]
+
+	if cmd == "" {
+		err = ErrCreateCommandtNotFoundOrEmpty
+		return
```
fabriciojs, line 42: no need to have this `return`

cmd/create.go
```diff
@@ -0,0 +1,80 @@
...
+	)
+
+	rootCmd.AddCommand(createCmd)
+}
+
+// NewKoolCreate creates a new handler for create logic
+func NewKoolCreate() *KoolCreate {
+	return &KoolCreate{
+		*newDefaultKoolService(),
+		&presets.DefaultParser{Presets: presets.GetAll()},
```
fabriciojs, line 33: I guess would be good if we could do this in a lazy fashion... Right now for absolutely every single execution of `kool` it will have to call NewKoolCreate on the `init()` of this file, and allocate the memory for the presets maps - which basically undo a change we made a while ago - https://github.com/kool-dev/kool/commit/788fc6b993b3cc8e173b961f60868911ac22c948

Not sure how to make it lazy as it is now though, need to think 🤔

cmd/create.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return
+}
+
+// NewCreateCommand initializes new kool create command
+func NewCreateCommand(create *KoolCreate) (createCmd *cobra.Command) {
+	createCmd = &cobra.Command{
+		Use:   "create [preset] [project]",
+		Short: "Create a new project using preset",
+		Args:  cobra.MinimumNArgs(2),
```
fabriciojs, line 75: we only expect the `2` arguments right? I guess we can make this `cobra.ExactNArgs` (not sure the validation rule name)

cmd/create.go
```diff
@@ -0,0 +1,80 @@
...
+		*NewKoolPreset(),
+	}
+}
+
+// Execute runs the create logic with incoming arguments.
+func (c *KoolCreate) Execute(originalArgs []string) (err error) {
+	preset := originalArgs[0]
+	dir := originalArgs[1]
+
+	if !c.parser.Exists(preset) {
```
fabriciojs, line 44: maybe we only instantiate the `c.parser` here, which would make it lazy enough already.

cmd/presets/parser.go
```diff
@@ -29,6 +30,21 @@ func (p *DefaultParser) Exists(preset string) (exists bool) {
 	return
 }
 
+// ErrCreateCommandtNotFoundOrEmpty error throwed when did not find the preset create command or it's empty
+var ErrCreateCommandtNotFoundOrEmpty = errors.New("create command not found or empty")
+
+// GetCreateCommand gets the command to create a new project
+func (p *DefaultParser) GetCreateCommand(preset string) (cmd string, err error) {
```
fabriciojs, line 37: just a thought for a later iteration - there might be some preset which requires more than a single command to "create" a new project... we probably could make this abstract, like we do for kool.yml `scripts:  task: single | multiline` do the same for the `preset_create`.

fabriciojs: And then here for example we could already return a Builder or something, to already encapsulate the runnable create commands.

cmd/presets/presets.go
```diff
@@ -193,6 +195,7 @@ networks:
 		".dockerignore": `/node_modules
 `,
 		"preset_language": "javascript",
+		"preset_create": "kooldev/node:14-nest nest new",
```
fabriciojs, line 198: Having the `preset_create` standard like this, making it mandatory to run only `kool docker` commands, I guess is an unnecessary constraint... why can't we simply leave it for the user the whole command? For example all these could simple be prefixed with `kool docker <image> <commands...>`, giving it more flexibility IMO.

fabriciojs (changes requested):
@leandroncbrito looking great! thanks for all the work on this!

I left a handful of comments, some we might leave for a feature iteration, others we may discuss further. Let me know what you think!

cmd/create.go
danielsuguimoto: @fabriciojs @leandroncbrito instead of adding the presets on `NewKoolCreate`, we can add the `LoadPresets` function and call it inside create command execution. The same can be done on preset/init

fabriciojs (approved):

PR #201, opened 2020-11-20
Fix current command Wordpress.md

renan-s-oliveira:
updated kool init wordpress to kool preset wordpress

### DIFF
--- a/docs/2-Presets/Wordpress.md
+++ b/docs/2-Presets/Wordpress.md
@@ -3,7 +3,7 @@
 If you want to create a project from scratch, just go to the directory you wanna have the new Wordpress install and run:
 
 ```bash
-$ kool init wordpress
+$ kool preset wordpress
 $ kool start
 ```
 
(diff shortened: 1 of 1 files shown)


docs/2-Presets/Wordpress.md
```diff
@@ -14,7 +14,7 @@ Upon first start, if the current working directory is not an existing Wordpress
 Go to the project folder and run:
 
 ```bash
-$ kool init wordpress
+$ kool preset wordpress
 ```
 
 **kool init** creates configuration files in order to enable you to configure and extend its behaviour.
```
danielsuguimoto, line 20: Please, could you change here as well?

danielsuguimoto (changes requested):
@renan-s-oliveira Thanks for your contribution! I've left just one note!

docs/2-Presets/Wordpress.md
renan-s-oliveira: ok.

danielsuguimoto (approved):
Thx!

PR #214, opened 2020-12-04
feat(run): allowing arguments in multiline scripts

### LINKED ISSUE #96: Proposal: Placeholders on kool.yml
Use placeholders on kool.yml for arguments, like a bash script? E.g.: `$1`, `$2` and so on...

This way we can use arguments in multi-lines kool scripts

Example:
```yml
script:
  - command1 $1
  - command2 $2 $3
```

Then we can call: `kool run script arg1 arg2 arg3`

danielsuguimoto:
Closes #96


dbpolito (changes requested):
actually chatting with Jun we considered moving from `--VAR=value` to `--set=ENV=value`  to be less magic and being able to document / add to help better...

fabriciojs:
I will add to the ticket a comment, this lacks design.

fabriciojs:
@danielsuguimoto given the history on this, the ongoing discussion on how to make scripts more flexible, and the understanding of using environment variables on the scripts to solve this injection, I guess we could close this one.

Let me know what you think!

danielsuguimoto:
@fabriciojs yes, you can close it...

Issue #224, opened 2020-12-15
Exclude fake components from build process


danielsuguimoto:
As discussed with the team, we don't need to compile the fake components, which are used only for testing, into the final binary files

fabriciojs:
Another idea, other than build tag conditionals, we could suffix files with _test, not sure about any possible downsides.

danielsuguimoto:
@fabriciojs build tags won't work at all, because the fake structs and interface are under the same package of the real ones, so the compiler understand that they must be available to entire package, so you cannot ignore them on building process. But we could use build tags if we put all tests and fakers in separated packages, keeping in the original packages only the code that will be built.

danielsuguimoto:
> Another idea, other than build tag conditionals, we could suffix files with _test, not sure about any possible downsides.

I'm not sure, but I guess that unnecessary code will be executed on our tests, if golang executes any function inside `*_test.go` files as test cases...

fabriciojs:
The functions executed should be only t ones prefixed with test.

Now about the build tags - you are saying we cannot ignore only one file
from one package? I didn’t get that but will follow up on research about it.

fabriciojs:
This is not something pressing, nor causing any noticeable downsides. Can we close this?

danielsuguimoto:
@fabriciojs yep, for me it's ok closing it...

PR #226, opened 2020-12-18
Preset Kool.yml customization

danielsuguimoto:
Branch based on #220

### DIFF
--- a/cmd/parser/fake_yml.go
+++ b/cmd/parser/fake_yml.go
@@ -1,0 +1,72 @@
+package parser
+
+import "kool-dev/kool/cmd/builder"
+
+// FakeKoolYaml implements all fake behaviors for using yml parser in tests.
+type FakeKoolYaml struct {
+	CalledParse, CalledHasScript, CalledParseCommands, CalledSetScript map[string]bool
+	CalledString                                                       bool
+	ScriptCommands                                                     map[string][]string
+	MockParseError                                                     map[string]error
+	MockHasScript                                                      map[string]bool
+	MockCommands                                                       map[string][]builder.Command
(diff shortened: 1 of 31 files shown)


cmd/parser/fake_yml.go
```diff
@@ -0,0 +1,72 @@
+package parser
```
fabriciojs, line 1: haven't we started adding `build ignore` tags to fake files?

cmd/parser/yml.go
```diff
@@ -38,6 +51,17 @@ func ParseKoolYaml(filePath string) (parsed *KoolYaml, err error) {
 	return
 }
 
+// Parse decodes the target kool.yml
+func (y *KoolYaml) Parse(filePath string) (err error) {
+	var parsed *KoolYaml
+	if parsed, err = ParseKoolYaml(filePath); err != nil {
+		return
```
fabriciojs, line 58: As Codecov complained, isn't there a test we can run with a bad formatted file, so we get this error covering hence the line?

fabriciojs (commented):
Looking very nice! Just a couple minor comments.

Although looking at all the changes here, I am starting to think: maybe we need to move the templates themselves off this repository? But that is a thought for the future...

cmd/parser/fake_yml.go
danielsuguimoto: we have a separated issue for that #224

cmd/parser/yml.go
danielsuguimoto: @fabriciojs yep, done!

fabriciojs (approved):

Issue #113, opened 2020-09-23
Kool deploy with files with accents


dbpolito:
If you have a file with accent like `importação` it will not be added on the `deploy.tgz` file.

PR #228, opened 2021-01-06
kool deploy - refactor + handling kool.deploy.env

### LINKED ISSUE #18: kool deploy - environment awarness
The `kool.deploy.env` file CANNOT be commited/versioned, but we should for adding it to the tarball.

fabriciojs:
Fixes #18 
Fixes https://github.com/kool-dev/kool/issues/113

- Refactor: Kool Service + Builder Commands. Still not being tested, but now a lot closer to being ready for doing so.

### DIFF
--- a/cmd/deploy.go
+++ b/cmd/deploy.go
@@ -3,11 +3,10 @@ package cmd
 import (
 	"fmt"
 	"kool-dev/kool/api"
-	"kool-dev/kool/cmd/shell"
+	"kool-dev/kool/cmd/builder"
 	"kool-dev/kool/environment"
 	"kool-dev/kool/tgz"
 	"os"
-	"os/exec"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


cmd/deploy.go
```diff
@@ -3,87 +3,116 @@ package cmd
...
+		Run:   DefaultCommandRunFunction(deploy),
+	}
+}
+
+// NewKoolDeploy creates a new pointer with default KoolDeploy service
+// dependencies.
+func NewKoolDeploy() *KoolDeploy {
+	defaultKoolService := newDefaultKoolService()
+	return &KoolDeploy{
+		*defaultKoolService,
```
danielsuguimoto, line 44:
```suggestion
		*newDefaultKoolService(),
```

cmd/deploy.go
```diff
@@ -3,87 +3,116 @@ package cmd
...
+	if url := d.envStorage.Get("KOOL_API_URL"); url != "" {
 		api.SetBaseURL(url)
 	}
 
-	fmt.Println("Create release file...")
-	filename, err = createReleaseFile()
+	d.Println("Create release file...")
+	filename, err = d.createReleaseFile()
 
 	if err != nil {
```
danielsuguimoto, line 69:
```suggestion
	if filename, err = d.createReleaseFile(); err != nil {
```

cmd/deploy.go
```diff
@@ -3,87 +3,116 @@ package cmd
...
 		}
 	}(filename)
 
 	deploy = api.NewDeploy(filename)
 
-	fmt.Println("Upload release file...")
+	d.Println("Upload release file...")
 	err = deploy.SendFile()
 
 	if err != nil {
```
danielsuguimoto, line 85:
```suggestion
	if err = deploy.SendFile(); err != nil {
```

cmd/deploy.go
```diff
@@ -131,44 +161,85 @@ func createReleaseFile() (filename string, err error) {
...
+	return
+}
 
+func (d *KoolDeploy) parseFilesListFromGIT(args []string) (files []string, err error) {
+	var output string
+
+	output, err = d.Exec(d.git, append([]string{"ls-files", "-z"}, args...)...)
+	if err != nil {
+		d.Error(err)
+		err = fmt.Errorf("failed listing GIT files")
```
danielsuguimoto, line 212: I think it should be better if we concat the error here instead of printing two messages

cmd/deploy.go
```diff
@@ -131,44 +161,85 @@ func createReleaseFile() (filename string, err error) {
...
+func (d *KoolDeploy) handleDeployEnv(files []string) []string {
+	path := filepath.Join(d.envStorage.Get("PWD"), koolDeployEnv)
+	if _, envErr := os.Stat(path); os.IsNotExist(envErr) {
+		return files
+	}
+
+	var isAlreadyIncluded bool = false
+	for _, file := range files {
+		if file == koolDeployEnv {
+			isAlreadyIncluded = true
```
danielsuguimoto, line 235: I guess if we return the files here, this variable will not be necessary and neither the if statement below

danielsuguimoto (changes requested):
Nice! Added just some small tweaks...

cmd/deploy.go
fabriciojs: I see, that is strictly correct. However, in this instance I rather keep the control variable... otherwise we mix-up the returning of `files` into an weird-loop structure that kind hides its original intent (simply check if the file already existed).

fabriciojs:
@danielsuguimoto PR updated.

danielsuguimoto (approved):

PR #231, opened 2021-01-11
kool deploy exec + refactoring to Kool Deploy API usage/lib client

fabriciojs:


api/deploy_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"kool-dev/kool/environment"
+	"testing"
+)
+
+func TestNewDeploy(t *testing.T) {
+	var tarball = "tarball"
+	d := NewDeploy(tarball)
+
+	if _, ok := d.env.(*environment.DefaultEnvStorage); !ok {
+		t.Errorf("unexpected default environment.EnvStorage")
```
danielsuguimoto, line 13: I have no idea how much extra work we have using `t.Errorf` instead of `t.Error`. In any case, I prefer using the last one when not parsing no variables into the error message.

fabriciojs:
@danielsuguimoto please can you review again?

We are fine with decreasing the test coverage at this point, we will soon focus on testing the whole API portion which should bring us close to complete coverage. But as this part is more prone to change, we can live with it as is now.

So if there are no other issues observed other than missing tests, please merge and we then proceed to tagging a new version!

danielsuguimoto (approved):

PR #232, opened 2021-01-12
Updates to Kool CLI Docs

indiegenus:


indiegenus:
@dbpolito @fabriciojs - free to edit the target and change to a different branch (that's not `master`). Also, not sure why my change set is showing as new files and not a diff. Thoughts?

dbpolito:
@indiegenus looks like you copied to the wrong folder...

Issue #241, opened 2021-01-22
Remove dependency on locally installed docker-compose


fabriciojs:
`docker-compose` has an official suggestion on using it from a container instead of installing the binary or python packages.

![image](https://user-images.githubusercontent.com/1116377/105420689-921c4580-5c1f-11eb-8964-7a5354ab14ef.png)

https://docs.docker.com/compose/install/#install-as-a-container  (go to "Alternative Install Options").

This would bring us a great benefit of not requiring users to install `docker-compose` (usually an extra step on setting up Docker locally).

Other than that, we would also pin down the version we use and remove unexpected behaviours.

---

In order to do that I guess we should mimic most of the official shell script into a `docker run` builder.Command internally.The affected areas would be:

```
cmd/start.go:42:                builder.NewCommand("docker-compose", "up", "-d", "--force-recreate"),
cmd/checker/checker.go:24:              builder.NewCommand("docker-compose", "ps"),
cmd/shell/shell.go:108: if exe == "docker-compose" {
cmd/shell/shell.go:139: if exe == "docker-compose" {
cmd/stop.go:40:         builder.NewCommand("docker-compose", "down"),
cmd/exec.go:41:         builder.NewCommand("docker-compose", "exec"),
cmd/logs.go:40:         builder.NewCommand("docker-compose", "ps", "-aq"),
cmd/logs.go:41:         builder.NewCommand("docker-compose", "logs"),
cmd/status.go:53:               builder.NewCommand("docker-compose", "ps", "--services"),
cmd/status.go:54:               builder.NewCommand("docker-compose", "ps", "-q"),
```

fabriciojs:
The script: https://github.com/docker/compose/releases/download/1.28.0/run.sh

fabriciojs:
Implemented, going in the next release.

PR #243, opened 2021-01-23
using docker-compose from a container when no docker-compose available

fabriciojs:
related to #241 #166

- [x] Refactor start, stop and exec commands
- [x] Refactor logs, status commands
- [x] Add tests to new `compose.DockerCompose` command
- [x] Fix always using --user on docker-compose exec
- [x] Make usage of local docker-compose when available as default
- [x] Update tests accordingly


cmd/exec.go
```diff
@@ -38,18 +41,30 @@ func NewKoolExec() *KoolExec {
...
 		&KoolExecFlags{false, []string{}, false},
 		environment.NewEnvStorage(),
-		builder.NewCommand("docker-compose", "exec"),
+		compose.NewDockerCompose("exec"),
 	}
 }
 
 // Execute runs the exec logic with incoming arguments.
 func (e *KoolExec) Execute(args []string) (err error) {
+	if asuser := e.env.Get("KOOL_ASUSER"); asuser != "" {
```
dbpolito, line 50: how much cost (time) it comes? maybe we could cache this somehow...

dbpolito (approved):

README.md
```diff
@@ -26,7 +26,7 @@ Get your local development environment up and running easy and quickly, put time
 
 ## Installation
 
-Kool is powered by [Docker](https://docs.docker.com/get-docker/) and [Docker Compose](https://docs.docker.com/compose/install/), you need to have them already installed on your machine.
+Requirements: Kool is powered by [Docker](https://docs.docker.com/get-docker/) so you need to have it already installed on your machine. If you still don't, please [get Docker first](https://docs.docker.com/get-docker/).
```
danielsuguimoto, line 29:
```suggestion
Requirements: Kool is powered by [Docker](https://docs.docker.com/get-docker/) so you need to have it already installed on your machine. If you haven't already, please [get Docker first](https://docs.docker.com/get-docker/).
```

docs/4-Commands/0-kool.md
```diff
@@ -21,7 +21,6 @@ Complete documentation is available at https://kool.dev/docs
 * [kool docker](kool-docker.md)	 - Creates a new container and runs the command in it.
 * [kool exec](kool-exec.md)	 - Execute a command within a running service container
 * [kool info](kool-info.md)	 - Prints out information about kool setup (like environment variables)
-* [kool init](kool-init.md)	 - [DEPRECATED] Proxies preset command
```
danielsuguimoto, line 24: this is being removed only on `2.0`... kool init is proxing kool preset and I guess the script to build de docs isn't handling it...

cmd/exec.go
danielsuguimoto: during a program execution, golang copies the entire environment variables list into its memory only once. So, I guess the only cost we have here is to lock the `map` and return the value.

danielsuguimoto (changes requested):

docs/4-Commands/0-kool.md
fabriciojs: not sure why this got removed, `make-docs ` made it... did you in the past manually returned this line? I guess deprecated commands/hidden commands do not show up by default in the extracted docs?

cmd/exec.go
fabriciojs: I believe Daniel misplaced the comment - the question was targeted at the `cat /etc/passwd` extra `exec` call that we will make now. And although we could think in some kind of cache, that would introduce the whole project-specific-storage of runtime configuration that we do not have currently, so I would leave with the overhead for now, and only consider this in the future.

docs/4-Commands/0-kool.md
danielsuguimoto: @fabriciojs have no idea why `make-docs` removes it.

cmd/exec.go
danielsuguimoto: @fabriciojs I got it! I agree that we should cache `passwd` as well. But for me is ok doing it in a next step.

docs/4-Commands/0-kool.md
fabriciojs: @danielsuguimoto just confirmed that if we remove the `initCmd.Deprecated` message from the `init.go` command, then `make-docs` will include it. `docs.go`, for the main command docs, do not include by default Non-available commands.

image

So since this is a builtin behavior, I am OK with sticking with it as well. The command is still available, and we do not show the deprecated commands in the list, just like `kool -h` do not:

```
➜  kool git:(docker-compose) kool
An easy and robust software development environment
tool helping you from project creation until deployment.
Complete documentation is available at https://kool.dev/docs

Usage:
  kool [command]

Available Commands:
  create      Create a new project using preset
  deploy      Deploys your application using Kool Dev
  docker      Creates a new container and runs the command in it.
  exec        Execute a command within a running service container
  help        Help about any command
  info        Prints out information about kool setup (like environment variables)
  logs        Displays log output from services.
  preset      Initialize kool preset in the current working directory. If no preset argument is specified you will be prompted to pick among the existing options.
  restart     Restart containers - the same as stop followed by start.
  run         Runs a custom command defined at kool.yaml in the working directory or in the kool folder of the user's home directory
  self-update Update kool to latest version
  start       Start the specified Kool environment containers. If no service is specified, start all.
  status      Shows the status for containers
  stop        Stop all running containers started with 'kool start' command

Flags:
  -h, --help      help for kool
      --verbose   increases output verbosity
  -v, --version   version for kool

Use "kool [command] --help" for more information about a command.
```

danielsuguimoto (approved):

PR #244, opened 2021-01-27
Hugo + update kool create

fabriciojs:
Adding new preset for Hugo the static site builder - https://gohugo.io/

-  :trophy: Feature - new preset

#### TL;DR

Adding new preset for Hugo to locally develop websites. Refactored how `preset-config.yml` `create` key handles the target folder.

- Add Hugo preset
- Changed `kool create <preset> <folder>` to put the `<folder>` into an environment variable and then updated the `create:` recipes on each existing preset to use it explicitly.

---

#### :dart: Kool Target version:

-  next **minor** release

#### Expected behavior

With the new implementation `kool create hugo my-website` should create a new Hugo site and install the presets; `kool preset hugo` should add both `kool.yml` and `docker-compose.yml` to the local directory.


danielsuguimoto (approved):

presets/hugo/kool.yml
```diff
@@ -0,0 +1,3 @@
+scripts:
+  hugo: kool docker -p 1313:1313 klakegg/hugo
```
danielsuguimoto, line 2: can't we use hugo entrypoint that we already have inside the container with `kool exec`?

fabriciojs: TBH I didn't give much thought into this. The good thing about using `kool docker` instead of `kool exec` is not depending on the running container. So the option would be `kool exec app hugo`. I am fine either way, let me know what you think is best!

dbpolito (approved):

PR #247, opened 2021-01-31
remove deploy file from laravel preset

fabriciojs:
Removing files from preset only meant for `kool deploy` which by itself doesn't make sense to have in the current default preset.

Later when we better structure and document the deploy feature we shall revisit such files.


dbpolito:
well while i agree the main idea was for using on kool cloud, i think it's still good for people that wants to deploy so their cloud env... i'm ok with removing as we don't have it documented... but we might want to re-introduce it even without kool cloud with some good docs...

fabriciojs:
@dbpolito I hear you! Indeed we can provide it as a helper for self-deploying. The idea in the back of my mind right now is introducing a secondary preset `laravel:deploy` so we have the necessary files. Let's put that in the roadmap soonish.

Issue #256, opened 2021-02-02
kool status - fatal error: concurrent map writes


fabriciojs:
**Describe the bug**

Sometimes` kool status` fails with what seems like a race condition on concurrent map writing.

```
$ kool status
fatal error: concurrent map writes
fatal error: concurrent map writes

goroutine 11 [running]:
runtime.throw(0xab8038, 0x15)
	/usr/local/go/src/runtime/panic.go:1116 +0x72 fp=0xc00006cc88 sp=0xc00006cc58 pc=0x4359f2
runtime.mapassign_faststr(0x9e80c0, 0xc000263b30, 0xab3d51, 0xe, 0x0)
	/usr/local/go/src/runtime/map_faststr.go:291 +0x3d8 fp=0xc00006ccf0 sp=0xc00006cc88 pc=0x4142d8
kool-dev/kool/cmd/shell.(*DefaultShell).LookPath(0xc00032e3c0, 0xb74780, 0xc0003035c0, 0x0, 0x0)
	/code/cmd/shell/shell.go:210 +0x13b fp=0xc00006cd50 sp=0xc00006ccf0 pc=0x85ccbb
kool-dev/kool/cmd/compose.(*DockerCompose).Args(0xc00032e410, 0xc0002237e0, 0x7f652d05c7d0, 0x0)
	/code/cmd/compose/docker-compose.go:59 +0x5e fp=0xc00006ce60 sp=0xc00006cd50 pc=0x85e73e
kool-dev/kool/cmd/shell.(*DefaultShell).Exec(0xc00032e230, 0xb747c0, 0xc00032e410, 0xc0002237e0, 0x1, 0x1, 0x0, 0x0, 0x0, 0x0)
	/code/cmd/shell/shell.go:104 +0x49 fp=0xc00006cef8 sp=0xc00006ce60 pc=0x85bf49
kool-dev/kool/cmd.(*DefaultKoolService).Exec(...)
	/code/cmd/kool_service.go:106
kool-dev/kool/cmd.(*KoolStatus).getServiceInfo(0xc000332000, 0xc0001fba6d, 0x5, 0xc000280a80, 0xc0001f7c30)
	/code/cmd/status.go:170 +0x13f fp=0xc00006cfb8 sp=0xc00006cef8 pc=0x95cfff
runtime.goexit()
	/usr/local/go/src/runtime/asm_amd64.s:1374 +0x1 fp=0xc00006cfc0 sp=0xc00006cfb8 pc=0x46a881
[...]

PR #257, opened 2021-02-02
Making shell.lookedUp cache thread-safe

fabriciojs:
Signed-off-by: fabriciojs <fajosesouza@gmail.com>

| Issue | #256 |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |

**Description**

We had a race condition that could generate a panic when writing simultaneously to a map (happened in some instances for `kool status`). The `shell.Exec` uses a cache for `os.LookPath` and such caching was unsafe for concurrency.

Introducing a new wrapper for the cache with `sync.RWMutext` for guaranteeing thread-safeness.

### DIFF
--- a/cmd/shell/lookupCache.go
+++ b/cmd/shell/lookupCache.go
@@ -1,0 +1,35 @@
+package shell
+
+import "sync"
+
+type lookupCache struct {
+	mtx   *sync.RWMutex
+	cache map[string]error
+}
+
+func newLookupCache() *lookupCache {
+	var mtx sync.RWMutex
+	return &lookupCache{&mtx, nil}
+}
+
+func (l *lookupCache) fetch(key string) (exists bool, err error) {
+	l.mtx.Lock()
+	defer l.mtx.Unlock()
+
+	if l.cache != nil {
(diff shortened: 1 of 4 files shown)


cmd/shell/lookupCache.go
```diff
@@ -0,0 +1,35 @@
...
+	cache map[string]error
+}
+
+func newLookupCache() *lookupCache {
+	var mtx sync.RWMutex
+	return &lookupCache{&mtx, nil}
+}
+
+func (l *lookupCache) fetch(key string) (exists bool, err error) {
+	l.mtx.Lock()
```
danielsuguimoto, line 16: Would `RLock()` and `RUnlock()`  be better? I guess it's not necessary locking the map for writing here

danielsuguimoto (changes requested):

cmd/shell/lookupCache.go
fabriciojs: @danielsuguimoto good catch! indeed had opted for RWMutex exactly because we can have a shared lock for reading.

danielsuguimoto (approved):

PR #265, opened 2021-02-04
multi input/output redirecting + go 1.15.7 upgrade

fabriciojs:
| Issue | #74  |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**

Input/Output redirection is now handled in a more flexible manner and it supports `kool.yml` like:

```yaml
scripts:
  cat: cat
  echo: echo
  out: kool run echo "content" > output
  in: kool run cat < output

  both: kool run in > output2

  bothatsametime: kool run cat < output > output3
``` 

Other than that, some tweaks were made to testing to fix some temporary bugs on caching. And more importantly:

- `go mod tidy` - cleared up go.mod for unnecessary stuff
- Go upgrade across the board to latest 1.15.7

---

**Notes**

There is a specific test case with the equivalent of the kool.yml sample above.


fabriciojs:
@danielsuguimoto ready for review - there are changes to test.yml workflow that should fix other branches failing after merge (letting go some bad caching for dependencies).

danielsuguimoto (approved):

Issue #251, opened 2021-02-01
Improve UX on typos for kool run


fabriciojs:
`kool run` will look for scripts with the given name on `kool.yml` files to execute. It is usual to mistype something.

Improvements could be:

- [x] Clearer error message, pointing to the similar/correct name.
- [x] Offer the option for the user to execute the corrected script (given it is under TTY and was found only one similar). In case of more than one similar, offer a wizard with the similars + "Cancel" option.

fabriciojs:
For the sake of history tracking, this issue recalls #207 although here we focus only on improving `kool run <script>` usage itself, and not falling back to it.

fabriciojs:
FInal part merged onto `master` and will be part of upcoming release.

PR #266, opened 2021-02-05
feat(run): prompt question about typo

danielsuguimoto:
| Issue | #251 |
| -----: | :-----: |
| :beetle: Bug Fix | No |
| :trophy: Feature | Yes |
| :pencil: Refactor | No |

**Description**

When a argument typo is detect, ask the user for the correct one, then executes it instead of the original argument.

---


cmd/root.go
```diff
@@ -79,8 +79,13 @@ func DefaultCommandRunFunction(services ...KoolService) CobraRunFN {
 			service.SetErrStream(cmd.ErrOrStderr())
 
 			if err := service.Execute(args); err != nil {
-				service.Error(err)
-				service.Exit(1)
+				if err.Error() == shell.ErrPromptSelectInterrupted.Error() {
```
fabriciojs, line 82: we should use more `errors.Is(err, targetErr)` for this type of checking

fabriciojs: other thing - for this to on the root command, I guess it should be more generic... maybe a new interface for `UserCancelled` implementing error coming from anywhere (in this case, answering No to the proposed fix).

fabriciojs (approved):

fabriciojs:
@danielsuguimoto good job! Thank you!

Issue #253, opened 2021-02-01
Proposal: mark/close stale issues/PRs


fabriciojs:
As part of ongoing improvements to project management and workflow, cleaning up old/abandoned stuff is a difficult but important part of life - we must let it go at some point.

The proposal here is to add [actions/stale](https://github.com/actions/stale) into our project so we mark either issues or pull requests as stale and then close them if still no activity is reported. Initial proposal:

- Make stale after 30 days
- Close after 10 more days after becoming stale

fabriciojs:
Merged onto master, will start cleaning up in no time :D

PR #272, opened 2021-02-13
add stale bot to housekeeping issues

fabriciojs:
| Issue | https://github.com/kool-dev/kool/issues/253 |
| -----: | :-----: |

**Description**

Adding https://probot.github.io/apps/stale/ to mark/close old issues without recent activities.

---

**Notes**

Overall we should benefit from a cleaner/simpler issues dashboard for bugs, questions, and external proposals only whereas we shall manage the backlog out of the default GH issues.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 15
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - R&D
+  - security
+  - enhancement
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,18 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 15
+# Issues with these labels will never be considered stale
+exemptLabels:
```
danielsuguimoto, line 6: should we add the `2.0` label here as well?

danielsuguimoto (changes requested):

.github/stale.yml
fabriciojs: @danielsuguimoto I am thinking no... my sense is that we should move to have only short-lived issues here on GH, and move the broader roadmap/improvements to internal trackers. This is based on the stats that the `/issues` page is the second most view page after the homepage, and the overwhelming number of issues and different topics there does not help in the first impression for newcomers. This is a proposal though of course.

danielsuguimoto: @fabriciojs okidoki!

danielsuguimoto (approved):

Issue #87, opened 2020-09-11
kool stop: Stop containers individually


danielsuguimoto:
Add the possibility of stoping containers individually. 

Ex.: `kool stop chromedriver` or `kool stop --services=chromedriver`

fabriciojs:
👍🏻  for the first suggestion, having the services as an argument list.

danielsuguimoto:
@fabriciojs then, should we change the `start` command as well? Today, it expects the `services` flag to specify which services it should start. So we can have both commands following the same pattern.

fabriciojs:
@danielsuguimoto I like and agree, we should change them both to be parameters list.

danielsuguimoto:
Start command changes made on #104

fabriciojs:
`docker-compose down` which is our default behaviour to map `kool stop` to, does not allow us to pick which containers to stop/remove.

![image](https://user-images.githubusercontent.com/1116377/95019325-13cd7080-063b-11eb-87d7-09e4b99a53a7.png)

Maybe we could fork the flow - if we specify services, we just `docker-compose stop` (which then must be incompatible with `--purge`), if no services are specified - we then default back to `docker-compose down`.

danielsuguimoto:
We can also use `docker-compose rm` command. It stops and removes the containers, but we won't be able to remove the related volumes, since it removes only the unnamed ones.

![image](https://user-images.githubusercontent.com/25229545/95084735-a9790680-06f4-11eb-829f-eb801f31e723.png)

PR #274, opened 2021-02-16
Kool stop - accepting service arguments

fabriciojs:
| Issue | #87  |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**

Added optional arguments to `kool stop` for stopping specific containers.

---

**Notes**

The behaviour is overall preserved where we destroy the container and, if `--purge`, also remove the named volumes.

### DIFF
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -65,7 +65,7 @@ func (s *KoolStop) Execute(args []string) (err error) {
 		if s.Flags.Purge {
 			s.rm.AppendArgs("-v") // removes volumes
 
-			s.Warning("Attention: when stopping specific services, only annonymous volumes will be removed.")
+			s.Warning("Attention: when stopping specific services, only anonymous volumes will be removed.")
 		}
 
 		s.rm.AppendArgs(args...)
--- a/cmd/stop.go
+++ b/cmd/stop.go
@@ -61,7 +61,7 @@ func (s *KoolStop) Execute(args []string) (err error) {
 		stopCommand = s.down
 	} else {
 		// we should only stop some services!
-		s.rm.AppendArgs("-s") // stops containers
+		s.rm.AppendArgs("-s", "-f") // stops containers; no interactive
 		if s.Flags.Purge {
 			s.rm.AppendArgs("-v") // removes volumes
 		}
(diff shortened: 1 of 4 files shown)


danielsuguimoto:
Reading the docs, a question came to my mind: does the `rm` command really removes named volumes?

image

fabriciojs:
@danielsuguimoto indeed, it won't erase named volumes as we use for database i.e.

I am not sure what to do. Should we warn about it? Should we error out as an incompatible combination of option and argument?

fabriciojs:
I am lining towards simply warning the user about not removing named volumes in the case of combining the argument and --purge option. What do you think?

danielsuguimoto:
@fabriciojs I'm ok about just warning. A note in the documentation would be nice as well. The advantage of stoping a single container is far more significant than the disadvantage of not removing the named volume, IMO. But I guess we can put a ticket for removing these volumes later in our backlog.

fabriciojs:
@danielsuguimoto As it seems, there is not a direct way for us to remove specific volumes created by `docker-compose` on `volumes:` section without `docker-compose down -v`. But agreed we should look into it in the future. Not sure if going straight to `docker volumes` is an option, and even if so, a smart one.

fabriciojs:
@danielsuguimoto warning added:

image

cmd/stop.go
```diff
@@ -39,29 +40,48 @@ func NewKoolStop() *KoolStop {
...
 }
 
 // NewStopCommand initializes new kool stop command
 func NewStopCommand(stop *KoolStop) (stopCmd *cobra.Command) {
 	stopCmd = &cobra.Command{
-		Use:   "stop",
-		Short: "Stop all running containers started with 'kool start' command",
-		Args:  cobra.NoArgs,
+		Use:   "stop [SERVICE]",
+		Short: "Stop and destroy running containers started with 'kool start' command. If not SERVICE is specified as argument all containers will be stopped.",
```
danielsuguimoto, line 84: I guess the correct is `If no SERVICE`, instead of not

danielsuguimoto (changes requested):
Found a typo. For me, after fixing it you can merge.

Issue #225, opened 2020-12-18
Live preview: "kool live" command for live public preview of local environment


fabriciojs:
**Describe the feature**

Would be awesome to have some feature for allowing easy sharing of local environment. Nowadays remote working reality makes  this an even more appealing feature. This is closely what `vagrant share` and others used to do.

Basically we need some server service for proxying/tunnelling the live requests to our local environment (having the tunnelling client up locally). Something like https://github.com/beyondcode/expose could be used so we don't need to write from scratch such solution.

- Having SSL enabled would be an even more awesome plus here, but that may depend on #28.
- The TTL for a "live" link should be relatively small. The server-side handling of this feature might go into kool.dev closed source (the most we can make open, the better), but we do need to have some control.

**Backward compatibility concerns**

No breaking changes. Adding a new command only, ideally not changing any others while doing so.

**Use cases**

- Developer might have some new UI they want to share with QA/PO for a quick validation. (very important por front-end developers).
- Developer might have access to an API and share it with a co-worker that for some reason cannot run such service, unblocking them.

**Examples**

```console
$ kool start
    ...
$ kool live
   .... creating your live preview link ...
   You can share the link to preview your local environment: https://preview-2311.kool.live
```

dbpolito:
@fabriciojs who is going to work on this one? assign and move to in progress when you define

fabriciojs:
We are going to use `kool share` to start, still open about what is the best term.

fabriciojs:
It is implemented and will be out in the next release as `kool share`

PR #275, opened 2021-02-17
added kool share command - make local environment available on kool.live

fabriciojs:
| Issue | #225  |
| -----: | :-----: |
| :trophy: Feature | Yes |
| :pencil: Refactor | Yes |

**Description**

Adding the new `kool share` command which allows for anyone using `kool` for local development to quickly share their local work through an HTTP tunnel on `kool.live` to anyone, anywhere. Think of it as `Ngrok` or `vagrant share` for `kool`!

---

**Notes**

- By default, the `app` service will be shared, but it can be changed using the `--service` flag.
- `kool.live` is a BeyondCode Expose server managed by us.

### DIFF
--- a/cmd/share.go
+++ b/cmd/share.go
@@ -14,6 +14,15 @@ import (
 type KoolShareFlags struct {
 	Service   string
 	Subdomain string
+	Port      uint
+}
+
+func (f *KoolShareFlags) parseServiceURI() string {
+	if f.Port != 0 {
+		return fmt.Sprintf("%s:%d", f.Service, f.Port)
+	}
+
+	return f.Service
 }
 
 // KoolShare holds handlers and functions to implement the stop command logic
@@ -41,7 +50,7 @@ func NewKoolShare() *KoolShare {
 	defaultKoolService := newDefaultKoolService()
 	return &KoolShare{
 		*defaultKoolService,
-		&KoolShareFlags{"app", ""},
+		&KoolShareFlags{"app", "", 0},
 		environment.NewEnvStorage(),
 		NewKoolStatus(),
 		builder.NewCommand("docker", "run", "--rm", "--init"),
@@ -68,7 +77,7 @@ func (s *KoolShare) Execute(args []string) (err error) {
 
 	s.share.AppendArgs("--network", s.env.Get("KOOL_GLOBAL_NETWORK"))
 	s.share.AppendArgs("beyondcodegmbh/expose-server:1.4.1", "share")
-	s.share.AppendArgs(s.Flags.Service)
+	s.share.AppendArgs(s.Flags.parseServiceURI())
 	s.share.AppendArgs("--server-host", "kool.live")
 
 	if s.Flags.Subdomain != "" {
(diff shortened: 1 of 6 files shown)


cmd/share.go
```diff
@@ -0,0 +1,98 @@
...
+		&KoolShareFlags{"app", ""},
+		environment.NewEnvStorage(),
+		NewKoolStatus(),
+		builder.NewCommand("docker", "run", "--rm", "--init"),
+	}
+}
+
+// validSubdomain runs the stop logic with incoming arguments.
+func (s *KoolShare) validSubdomain(subdomain string) bool {
+	return regexp.MustCompile("^[a-z]+[a-z0-9]+$").MatchString(subdomain)
```
dbpolito, line 52: why this logic / validation? can't have _ or -?

cmd/share.go
```diff
@@ -0,0 +1,98 @@
...
+// NewShareCommand initializes new kool stop command
+func NewShareCommand(share *KoolShare) (shareCmd *cobra.Command) {
+	shareCmd = &cobra.Command{
+		Use:   "share",
+		Short: "Live share your local environment through an HTTP tunnel with anyone, anywhere.",
+		Args:  cobra.NoArgs,
+		Run:   DefaultCommandRunFunction(share),
+	}
+
+	shareCmd.Flags().StringVarP(&share.Flags.Service, "service", "", "app", "The name of the local service container we want to share.")
```
dbpolito, line 95: shall we add port as separate argument? services like next / nuxt uses port 3000, or we just do: `--service=app:3000`?

dbpolito: also thinking about protocol, if you use ssl locally (when we hopefully properly support it) you might want to tunnel to https://somethingelse... having this param as `--service` may be weird...

cmd/share.go
fabriciojs: yeah this needs improvement, I will get thee RFC regex

cmd/share.go
fabriciojs: Indeed the `port` might be a problem... I wouldn't worry now with the protocol thing

I will add a `--port` option, which defaults to the first published port from the running container.

dbpolito: well we can leave port empty as default... and the user provide it when necessary... we don't have to be that smart at the beginning

fabriciojs: empty means default to `80` right? I can live with that

fabriciojs:
@dbpolito PR updated!

docs/4-Commands/kool-share.md
```diff
@@ -0,0 +1,26 @@
...
+
+```
+kool share [flags]
+```
+
+### Options
+
+```
+  -h, --help               help for share
+      --service string     The name of the local service container we want to share. (default "app")
```
dbpolito, line 13: need to be regenerated to have new port option

dbpolito (approved):

danielsuguimoto (approved):

PR #280, opened 2021-03-04
Improving database templates

dbpolito:
| Issue | # |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |
| :trophy: Feature | Yes |
| :pencil: Refactor | Yes |

**Description**

This fixes a bug when you had `DB_PASSWORD` empty, like in laravel default setup... with this fix it will just work with laravel's default env, being `DB_USERNAME=root` and `DB_PASSWORD=`...

On bash `${DB_PASSWORD:-pass}` will check if var exists and not empty, so if you had it empty it would still go to fallback pass, now `${DB_PASSWORD-pass}` only check if var doesn't exist... so if it's empty it will respect that.

Added `healthcheck`.

Also did some improvements to scripts and formatting.

### DIFF
--- a/cmd/presets/templates.go
+++ b/cmd/presets/templates.go
@@ -68,6 +68,8 @@ volumes:
       - cache:/data:delegated
     networks:
       - kool_local
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
 
 volumes:
   cache:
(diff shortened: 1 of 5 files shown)


indiegenus (approved):

indiegenus:
@dbpolito - nice work. Looks good to me.

danielsuguimoto (approved):

PR #285, opened 2021-03-08
Updates to Command Docs

indiegenus:
| Issue | # |
| -----: | :-----: |
| :beetle: Bug Fix | No |
| :trophy: Feature | No |
| :pencil: Refactor | No |

**Description**

- [ ] updated short and long descriptions for Commands, as well as formatting of flags (where appropriate)
- [ ] made consistent use of the `...` notation (i.e. `[service...]` and `[option...]`) to show acceptance of multiple arguments
- [ ] made the long description for `completion.go` match the recently updated description used in the Autocompletion section of [the Installation doc](https://kool.dev/docs/getting-started/installation) page

---

**Notes**

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -93,8 +93,8 @@ func (r *KoolRun) Execute(originalArgs []string) (err error) {
 // NewRunCommand initializes new kool stop command
 func NewRunCommand(run *KoolRun) (runCmd *cobra.Command) {
 	runCmd = &cobra.Command{
-		Use:   "run [command]",
-		Short: "Run the specified [command], which are defined as scripts in a kool.yml file located in the working directory or in the kool folder of the user's home directory.",
+		Use:   "run [script] [command]",
+		Short: "Run the specified [script] (as defined in the kool.yml file). For single-line scripts, run the [script] with an optional [command].",
 		Args:  cobra.MinimumNArgs(1),
 		Run:   DefaultCommandRunFunction(run),
 		ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -93,8 +93,8 @@ func (r *KoolRun) Execute(originalArgs []string) (err error) {
 // NewRunCommand initializes new kool stop command
 func NewRunCommand(run *KoolRun) (runCmd *cobra.Command) {
 	runCmd = &cobra.Command{
-		Use:   "run [SCRIPT]",
-		Short: "Runs a custom command defined at kool.yaml in the working directory or in the kool folder of the user's home directory",
+		Use:   "run [script]",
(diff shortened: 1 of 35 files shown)


indiegenus:
@dbpolito - FYI - I did _not_ include an update for the Laravel preset doc page in this PR because I want to think a little more about the Presets. A straight copy/paste would create duplicate content (as determined by search engines), which we want to avoid. This is usually handled by adding a canonical tag in the `<head>` section, but I don't think we have that ability here. Is that true?

dbpolito:
Well not at the moment, we can bring the idea of meta to introduce support like other CMS or maybe it's a indication we need to  move to a more robust tool like hugo or gitbook

dbpolito (approved):

cmd/run.go
```diff
@@ -93,8 +93,8 @@ func (r *KoolRun) Execute(originalArgs []string) (err error) {
 // NewRunCommand initializes new kool stop command
 func NewRunCommand(run *KoolRun) (runCmd *cobra.Command) {
 	runCmd = &cobra.Command{
-		Use:   "run [SCRIPT]",
-		Short: "Runs a custom command defined at kool.yaml in the working directory or in the kool folder of the user's home directory",
+		Use:   "run [script]",
```
dbpolito, line 96: maybe add `[args...]` here and mention the difference between single / multi line commands?

dbpolito: at `exec` command we call this just `[command]`: https://github.com/kool-dev/kool/pull/285/files#diff-ce92eefe0908ac93cdc54bd33b7d01c347c159ae09a9324bd0479acdbc38c442R88

maybe better think and follow the standard.

dbpolito (changes requested):
a few notes

danielsuguimoto (commented):
Every time you change to documentation, you can run `kool run make-docs` and add the updated markdown files to the PR. Thus, we will not need to do this before the release or take the risk of releasing the code without updating the markdown files. Thanks!

cmd/run.go
indiegenus: @dbpolito - I haven't yet addressed your comment about `[args...]` and single/multi line commands. But I did replace `run [script]` with `run [command]`, for consistency. Good catch!

dbpolito:
Not sure changing to command is better... At exec we have service and command, which was the args a mentioned, which is forwarded to the container.

Now on run, we have a script under scripts key, so following the standard would be [script] [command], where command is optional and only relevant for single line scripts.

indiegenus:
I misunderstood your comment, and made a hasty update. I’ll revert.

indiegenus:
@dbpolito - I just submitted https://github.com/kool-dev/kool/pull/285/commits/f4f16f2c03fc574d9903c23f87bb700fb59fbaca as a fix to the `kool run` command.

dbpolito (approved):

danielsuguimoto (approved):

PR #292, opened 2021-03-18
Environment Variable flags to run command

danielsuguimoto:
| Issue | # |
| -----: | :-----: |
| :beetle: Bug Fix | No |
| :trophy: Feature | Yes|
| :pencil: Refactor | No |

**Description**
This Pull Request adds a new flag that allows `kool run` command to set environment variables at runtime

---

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -31,9 +31,6 @@ var ErrExtraArguments = errors.New("error: you cannot pass in extra arguments to
 // ErrKoolScriptNotFound means that the given script was not found
 var ErrKoolScriptNotFound = errors.New("script was not found in any kool.yml file")
 
-// ErrExtraEnvMultiLines Extra environment variables error
-var ErrExtraEnvMultiLines = errors.New("error: you cannot pass in extra environment variables to multiple commands scripts")
-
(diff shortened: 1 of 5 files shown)


cmd/run.go
```diff
@@ -78,6 +82,11 @@ func (r *KoolRun) Execute(originalArgs []string) (err error) {
 		return
 	}
 
+	if len(r.Flags.EnvVariables) > 0 && len(r.commands) > 1 {
+		err = ErrExtraEnvMultiLines
```
fabriciojs, line 86: we do error out when passing extra parameters to multi-line scripts... because parameters are most certainly signature aware so having them for multi-line script is assumed to be a mistake and possible source of problems...

but environment variables on the other hand do not affect the commands necessarily. So I guess for environment variables we are good to have them for all commands in a multi-line script.

fabriciojs (changes requested):
@danielsuguimoto all in all looks good! Just one comment about not restricting this to single-line scripts.

Beyond that, just curious - how did this come to be? Is this part of something else, some new use case, or a path to another implementation?

dbpolito (approved):

fabriciojs (approved):

Issue #186, opened 2020-10-27
Warn about development kool version


danielsuguimoto:
**Describe the feature**

When you run `kool run dev`, the installed `kool` command is replaced by the one built locally. So the user can run it and forget to update `kool` back to a production version. It would be nice to warn the users that they are using a development version.

**Backward compatibility concerns**

Would this be BC? No

**Use cases**

The users are also `kool` contributors and run `kool run dev` to test a feature that they are developing and they forget to update `kool` back to a production version. If they start working on other projects with `kool` later, they might face some instabilities.

fabriciojs:
I like this! I guess we can make this safe with:

1. Change the default variable value to a constant, so we can later check whether the variable value is equal to such constant (meaning it is a development build).
2. On the root command PreRunPersistent we put a new service for this validation.
3. In such service, we check:
     - Are we running on a TTY? (I guess we should only warn about on TTY, so we know nothing will "break" inadvertently due to the output)
     - Is this a development build?
     - Now comes the trick part... for tests for example, the version will be development build - should we change it somehow (don't know if possible) to a testing build version?

---

After enumerating the items above I come to a different idea - what if instead of "warning" the user he/she is using a development build *after* they use it, we change the `kool run dev` to instead of override `kool` from PATH, to actually add a new `koold` - standing for `kool development build`, so the user still have original kool, and whenever they want they can use `koold` for testing the just built environment.

For debug purposes as well, maybe for this development build we could also attach to the version the short git commit, just so users can tell what development build they are running on exactly.

I am tending towards this second idea now that it came to light. Let's hear other thoughts!

danielsuguimoto:
Sometimes I like to override the original `kool` command, sometimes I don't. I usually like to replace it when I need to test the integration between the `kool run` command and any command without having to change `kool.yml` to do this.

fabriciojs:
That is a good point. What if, under a development build, we check for existing `koold`, and if so, replace any `kool` on `kool run` scripts to `koold`? Too hacky?

fabriciojs:
Let's add the warning only.

fabriciojs:
This has a merged implementation, will go into the next release 1.11.0

PR #295, opened 2021-03-27
warning about development build

fabriciojs:
| Issue | #186  |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**

We have a new warning for every time you run `kool` in a development build. This should help prevent inadvertent usage of edge builds for Kool development team on other projects.

---

**Notes**

- The warning is only issued when output is a TTY (so we don't create hard to debug situations).
- The output is issues only once (even for recursive calls)


fabriciojs:
Closes v

dbpolito (approved):

PR #296, opened 2021-03-28
.env.local load + docs update

fabriciojs:
| Issue | - |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**

This PR introduces the preferred loading of `.env.local` before `.env`. It also updates documentation to cover that and improves some overall language.

---

**Notes**

The `.env.local` usage is a common pattern across some *dot env* powered frameworks. This style of usage suggests versioning `.env` while having `.env.local` non-versioned and used for tweaking specific environments. This is true for Symfony.

*There could be some drawback* on this for projects using .env.local with kool, where you might see some behavior drift. I don't believe though that is cause to hold this to 2.0, I would ship now with just an alert on the changelog.


fabriciojs:
@indiegenus asking for your review mainly due to the documentation changes.

dbpolito:
I'm sure i link moving into this direction... One point is that laravel does support this kind of env, but i'm not sure if also does merge or just takes precedence (i think it just takes precedence).

I'm not against it, but never been a huge fan of this approach.

danielsuguimoto (changes requested):
We should add `.env.local` to our `.gitignore`, I guess...

danielsuguimoto:
@dbpolito @fabriciojs what about adding a global flag `--env`, so we can force kool to load only a specific env file?

danielsuguimoto:
@fabriciojs for example, laravel environments...we have `.env` for a development environment and `.env.testing` for a testing one. Usually on database scripts, we need to hardcode the testing database name...

indiegenus (changes requested):
@fabriciojs - I submitted my edits in https://github.com/kool-dev/kool/pull/301

environment/env_default.go
```diff
@@ -1,19 +0,0 @@
...
-# Only print out the generated commands instead of running them
-# KOOL_DEBUG=0
-# Prints out all commands and their output
-# KOOL_VERBOSE=0
-
-# Optional, default is folder name
-# KOOL_NAME=custom_name
-
-# Host user for mapping execution within the containers using kool images
-# KOOL_ASUSER=$UID
```
indiegenus, line 15: @fabriciojs @dbpolito - does the elimination of these defaults require us to change the Docs in any way (as far as how users set up their `.env` file)?

fabriciojs: @indiegenus it does not! all of those defaults were commented (` # `), they actually had no effect at all.

fabriciojs:
> @fabriciojs for example, laravel environments...we have `.env` for a development environment and `.env.testing` for a testing one. Usually on database scripts, we need to hardcode the testing database name...

@danielsuguimoto so this is not so simple... because even though we would load the environment variable, this could mislead people in believing their application would also use that. (i.e using .env.testing on Kool won't necessarily change Laravel behaviour... even though it may work right now, I guess it is risky to assume it would for all usages).

danielsuguimoto (approved):

danielsuguimoto:
@fabriciojs I understand your concern...well, it was just an idea...no need to discuss about or implement it here...this PR is ok as it is now IMO

PR #302, opened 2021-04-06
Docker Run With Network

fbrissi:
| Issue | - |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**
There are chances that I need to run a docker with a network.   
An example when I need create a docker script that execute `kool docker` and use an servico in `docker-compose`

---

**Notes**
I add an condicional to run **Sync Docker Hub Description** only branch **master** and only repository `kool-dev/kool`

### DIFF
--- a/.github/workflows/docker-description.yml
+++ b/.github/workflows/docker-description.yml
@@ -15,6 +15,7 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Sync Docker Hub Description
+        if: github.ref == 'refs/heads/master' && github.repository == 'kool-dev/kool'
         uses: peter-evans/dockerhub-description@v2
         env:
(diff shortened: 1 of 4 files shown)


fabriciojs (approved):
@fbrissi thanks for the PR!

This implementation is fine and I think it is good to have it as an explicit flag.

But just so we are clear - did you know that we can already to this by escaping the options and passing them down to `docker run` ?

Right now the way to do it is: `kool docker -- --network my-network image:tag cmd ...`

This way with `--` you tell kool to stop parsing options/flags and already receive the all following arguments as literal arguments, which will make all of them pass down to `docker run` in the end.

dbpolito:
I agree, maybe the most used ones like network and user worth mapping

dbpolito (approved):

Issue #316, opened 2021-05-01
Permission Problems with Hugo Preset on Linux


indiegenus:
**Describe the bug**

This is based on a development version of `kool`, based on WIP refactoring of the Presets.

Uses Docker Image [`klakegg/hugo`](https://hub.docker.com/r/klakegg/hugo/).

Unable to complete `kool create` because of permission issue (see "error: failed to write preset file docker-compose.yml: open docker-compose.yml: permission denied").

```bash
current-user /home/git
bsh ❯ kool create hugo my-project
Warning: you are executing a development version of kool.
Congratulations! Your new Hugo site is created in /app/my-project.
Just a few more steps and you're ready to go:
1. Download a theme into the same-named folder.
   Choose a theme from https://themes.gohugo.io/ or
   create your own with the "hugo new theme <THEMENAME>" command.
2. Perhaps you want to add some content. You can add single files
   with "hugo new <SECTIONNAME>/<FILENAME>.<FORMAT>".
3. Start the built-in live server via "hugo server".
Visit https://gohugo.io/ for quickstart guide and full documentation.
Preset hugo is initializing!
error: failed to write preset file docker-compose.yml: open docker-compose.yml: permission denied
```

The `my-project` directory is created a `root:root`.

```bash
current-user /home/git
bsh ❯ ls -l
total 24
drwxr-xr-x  8 root root 4096 May  1 09:00 my-project
```

**Kool version and environment**

Kool 1.11.0
Linux (Linux Mint 20 - Ubuntu 20.04)

**To Reproduce**

- `kool create hugo my-project`

**Expected behavior**

The project directory ($CREATE_DIRECTORY) is created for the current user, not the `root` user.

```bash
current-user /home/git
bsh ❯ ls -l
total 24
drwxr-xr-x  8 current-user current-user 4096 May  1 09:00 my-project
```

indiegenus:
@fabriciojs - I'm also experiencing a permissions problem on Linux with the WordPress preset. After running `kool start` using Kool's WordPress Docker image, the WordPress files are created with `1337` as the `user:group`.

```bash
swp /home/git/my-project via 🐘
bsh ❯ ls -la
total 244
drwxrwxr-x  6 swp  swp   4096 May  2 05:27 .
drwxr-xr-x  8 swp  swp   4096 May  1 15:00 ..
-rwxrwxr-x  1 swp  swp   1136 May  1 15:00 docker-compose.yml
[...]

PR #322, opened 2021-05-08
Updates to wordpress preset

fabriciojs:
| Issue | #316  |
| -----: | :-----: |

**Description**

Added new PHP 8.0 option for Wordpress preset.

### DIFF
--- a/cmd/presets/configs.go
+++ b/cmd/presets/configs.go
@@ -300,11 +300,13 @@ commands:
 questions:
   compose:
     - key: app
-      default_answer: PHP 7.4
+      default_answer: PHP 8.0
       message: What app service do you want to use
       options:
         - name: PHP 7.4
           template: wordpress74.yml
+        - name: PHP 8.0
+          template: wordpress80.yml
     - key: database
       default_answer: MySQL 5.7
       message: What database service do you want to use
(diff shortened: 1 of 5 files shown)


fabriciojs:
@kool-dev/core-members @indiegenus - the `wordpress` preset currently have questions for `cache` (Redis/Memcached) and Node package (yarn/npm)...

- Does it really make sense to offer upfront such cache service? Do we need to add further documentation on how to use either of the two options (changing on wp-config.php).

- Does it make any sense to have yarn/npm/npx as default scripts on a Wordpress install? How rare is someone would use it?

dbpolito:
i think we decided to remove npm/yarn on our call, maybe forgot to remove... cache i think it's good to keep... there are configs that doesn't require code changes, just config... W3TotalCache does. but it's very straightforward and they do have good docs, so nothing needed in our end.

Maybe just point what REDIS_HOST will look like

cmd/presets/configs.go
```diff
@@ -300,11 +300,13 @@ commands:
 questions:
   compose:
     - key: app
-      default_answer: PHP 7.4
+      default_answer: PHP 8.0
       message: What app service do you want to use
```
indiegenus, line 304: @fabriciojs - this does not appear to be the latest parsing of the presets, because the `message` should have been changed to `message: What PHP version do you want to use`.

Issue #319, opened 2021-05-05
Error when try to run without docker-compose installed


billlima:
```
$ kool run composer install

$ (TTY in: true out: true) kool [exec app composer install]
[recursive call]
[...]

PR #323, opened 2021-05-09
fix docker-compose container version TTY carring over

fabriciojs:
| Issue | https://github.com/kool-dev/kool/issues/319 |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |

**Description**

Fixes carrying over TTY awareness for docker-compose command when using container strategy.

### DIFF
--- a/cmd/compose/docker-compose.go
+++ b/cmd/compose/docker-compose.go
@@ -12,6 +12,11 @@ import (
 // DockerComposeImage holds the Docker image:tag to use for Docker Compose
 const DockerComposeImage = "docker/compose:1.28.0"
 
+// TtyAware interface holds functions for becoming aware of TTY
+type TtyAware interface {
+	SetIsTTY(bool)
+}
+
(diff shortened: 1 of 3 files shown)


cmd/compose/docker-compose.go
```diff
@@ -12,6 +12,11 @@ import (
 // DockerComposeImage holds the Docker image:tag to use for Docker Compose
 const DockerComposeImage = "docker/compose:1.28.0"
 
+// TtyAware interface holds functions for becoming aware of TTY
+type TtyAware interface {
```
danielsuguimoto, line 16: why do we need a separated interface for it?

fabriciojs: so we can have a generic check - the conditional for knowing when to inject it... before it was done using the struct pointer itself. This way, with the interface, we can mock/replace during tests.

I started but aborted writing such tests, but kept this part as it is a step closer on that direction.

PR #343, opened 2021-05-28
fix JS package manager on PHP presets

fabriciojs:
| Issue | - |
| -----: | :-----: |
| :beetle: Bug Fix | Yes|

**Description**

Fixes the npm/yarn scripts for Laravel and Symfony presets. Using `kool docker` instead of the bad `kool exec`.

### DIFF
--- a/core/presets/configs.go
+++ b/core/presets/configs.go
@@ -112,9 +112,9 @@ questions:
       message: What javascript package manager do you want to use
       options:
         - name: npm
-          template: npm-laravel.yml
+          template: npm.yml
         - name: yarn
-          template: yarn-laravel.yml
+          template: yarn.yml
 templates:
   - key: scripts
     template: laravel.yml
--- a/templates/scripts/npm.yml
+++ b/templates/scripts/npm.yml
@@ -3,4 +3,4 @@ scripts:
   npx: kool docker kooldev/node:14 npx
   node-setup:
     - kool run npm install
-    - kool run npm dev
+    - kool run npm run dev
--- a/templates/scripts/npm.yml
+++ b/templates/scripts/npm.yml
@@ -1,3 +1,3 @@
 scripts:
-  npm: kool exec app npm
-  npx: kool exec app npx
+  npm: kool docker kooldev/node:14 npm
+  npx: kool docker kooldev/node:14 npx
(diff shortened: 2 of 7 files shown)


danielsuguimoto (commented):
both `yarn-laravel.yml` and `npm-laravel.yml` have the script `node-setup`, but `npm.yml` and `yarn.yml` don't and laravel preset uses this script in its `setup` script. Isn't these changes causing an error in laravel setup?

core/presets/configs.go
```diff
@@ -112,9 +112,9 @@ questions:
       message: What javascript package manager do you want to use
       options:
         - name: npm
-          template: npm-laravel.yml
```
indiegenus, line 115: @fabriciojs - the reason we have `-laravel` versions of the `npm` and `yarn` scripts is because the `laravel.yml` template is the only one that requires a `node-setup` script (which is different for `npm` vs. `yarn`). Based on these changes, the `setup` and `reset` scripts in `kool.yml` are now broken.

https://github.com/kool-dev/kool/blob/4e159861d4b8ff1f14910787e8c49ca27750531b/templates/scripts/laravel.yml#L11

https://github.com/kool-dev/kool/blob/4e159861d4b8ff1f14910787e8c49ca27750531b/templates/scripts/laravel.yml#L16

indiegenus (changes requested):

danielsuguimoto:
IMO, there is no problem adding `node-setup` in both `npm.yml` and `yarn.yml`

fabriciojs:
Thanks @danielsuguimoto and @indiegenus ! 

PR has been updated. I decided to follow Jun's suggestion and added `node-setup` on `npm'/`yarn' templates. The side effect is that other `presets` could have it now, even though they may not call `node-setup` on their own `setup` script. Which is OK since the preset does have a package manager, now it will have the `node-setup` script available (i.e the Symfony preset falls into this case).

templates/scripts/npm.yml
```diff
@@ -1,3 +1,6 @@
 scripts:
-  npm: kool exec app npm
-  npx: kool exec app npx
+  npm: kool docker kooldev/node:14 npm
+  npx: kool docker kooldev/node:14 npx
+  node-setup:
+    - kool run npm install
+    - kool run npm dev
```
danielsuguimoto, line 6: I'm not 100% sure, but unlike `yarn`, `npm` needs the `run` command explicitly

danielsuguimoto (changes requested):

danielsuguimoto (approved):

PR #359, opened 2021-07-08
MySQL root user error troubleshooting on docs

renancaride:
| Issue | # |
| -----: | :-----: |
| :beetle: Bug Fix | No |
| :trophy: Feature | No |
| :pencil: Refactor | No |

**Description**
Adding a troubleshooting page on docs to handle the mysql access denied error when the username is set to root on env files

### DIFF
--- a/docs/2-Presets/Laravel.md
+++ b/docs/2-Presets/Laravel.md
@@ -97,6 +97,16 @@ PostgreSQL 13.0
 +DB_PORT=5432
 ```
 
+> In order to avoid permission issues with mysql, add a user other than root and a password to your **.env.example** file
+
+```diff
+-DB_USERNAME=root
++DB_USERNAME=<some_user>
+
+-DB_PASSWORD=
++DB_PASSWORD=<somepass>
+```
+
 ### Cache Services
 
 Redis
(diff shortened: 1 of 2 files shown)


docs/2-Presets/Laravel.md
```diff
@@ -97,6 +97,16 @@ PostgreSQL 13.0
 +DB_PORT=5432
 ```
 
+> In order to avoid permission issues with mysql, add a user other than root and a password to your **.env.example** file
```
danielsuguimoto, line 100: I'm not sure if this is an issue that we can face only with mysql, so I guess this text can be more generic.

docs/6-Troubleshooting/mysql-perm-error.md
```diff
@@ -0,0 +1,35 @@
+### `kool run mysql` or `kool run setup`
+
+**Problem:**
+
+> Access denied for root on mysql
```
danielsuguimoto, line 5: Same as previous comment...if it is not a mysql issue only, the text could be more generic

danielsuguimoto (changes requested):

docs/2-Presets/Laravel.md
renancaride: Just confirmed on the other presets available, it only happens on MySQL and mariaDB. I've updated the docs.

danielsuguimoto (approved):

PR #365, opened 2021-08-09
Fix input forwarding conflict

fabriciojs:
| Issue | https://github.com/kool-dev/kool/pull/364 |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |

**Description**

This fixes a bug introduced on https://github.com/kool-dev/kool/pull/364, while maintaining the original fix as well.

The correction of the check for user on running container when non-TTY env exposed another issue, where the re-use of the same file descriptors for an extra exec was causing for conflict when on that same non-TTY env you were trying to use input redirection from a file.


fabriciojs:
TODO: add a regression test.

Issue #391, opened 2021-11-17
Using Profiles With Compose 


fbrissi:
I want to [Use Profiles With Compose](https://docs.docker.com/compose/profiles/), that would be very useful:

![image](https://user-images.githubusercontent.com/339706/142203144-1eee54c8-822a-4491-98d4-9ae84ceffb8b.png)

PR #392, opened 2021-11-17
Support to use docker-compose profile

fbrissi:
| Issue | [#](https://github.com/kool-dev/kool/issues/391)|
| -----: | :-----: |
| :beetle: Bug Fix | No |
| :trophy: Feature | Yes |
| :pencil: Refactor | No |

**Description**
Add `--profile` arg to `kool start` to [Use Profiles With Compose](https://docs.docker.com/compose/profiles/)

---

**Notes**
 - I don't know how we can create a test for this case, if someone can help it would be good.

### DIFF
--- a/commands/start.go
+++ b/commands/start.go
@@ -15,6 +15,7 @@ import (
 type KoolStartFlags struct {
 	Foreground bool
 	Rebuild    bool
+	Profile    string
 }
 
 // KoolStart holds handlers and functions for starting containers logic
@@ -51,6 +52,7 @@ all containers are started. If the containers are already running, they are recr
 
 	startCmd.Flags().BoolVarP(&start.Flags.Foreground, "foreground", "f", false, "Start containers in foreground mode")
 	startCmd.Flags().BoolVarP(&start.Flags.Rebuild, "rebuild", "b", false, "Updates and builds service's images")
+	startCmd.Flags().StringVarP(&start.Flags.Profile, "profile", "", "", "Specify a profile to enable")
 
 	return
 }
@@ -61,7 +63,7 @@ func NewKoolStart() *KoolStart {
 	defaultKoolService := newDefaultKoolService()
 	return &KoolStart{
 		*defaultKoolService,
-		&KoolStartFlags{false, false},
+		&KoolStartFlags{false, false, ""},
 		checker.NewChecker(defaultKoolService.shell),
 		network.NewHandler(defaultKoolService.shell),
 		environment.NewEnvStorage(),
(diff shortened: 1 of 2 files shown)


danielsuguimoto:
@fbrissi for testing it...you only need to create a test case to check if the flags were added to the `docker-compose` internal command

danielsuguimoto:
@fbrissi kool is prepared to handle commands with the following structure:
- `[docker/docker-compose] [command] [arguments]`

But this PR is trying to make the following structure to work:
- [docker/docker-compose] [options] [command] [arguments]`

@fabriciojs do you have any thoughts? I was thinking about adding another array to command struct...other than Cmd and Args...to handle these options...

commands/start.go
```diff
@@ -96,6 +98,13 @@ func (s *KoolStart) Execute(args []string) (err error) {
 		}
 	}
 
+	if len(s.Flags.Profile) > 0 {
+		if aware, ok := s.start.(compose.LocalDockerComposeAware); ok {
```
fabriciojs, line 102: Why have you felt the need to introduce this new interface? why not simply call `s.start.AppendArgs("--profile", s.Flags.Profile)`?

fabriciojs:
> @fbrissi kool is prepared to handle commands with the following structure:
> 
> * `[docker/docker-compose] [command] [arguments]`
> 
> But this PR is trying to make the following structure to work:
> 
> * `[docker/docker-compose] [options] [command] [arguments]`. Ex.: `docker-compose --profile profile-1 up ...`
> 
> @fabriciojs do you have any thoughts? I was thinking about adding another array to command struct...other than Cmd and Args...to handle these options...

@danielsuguimoto I believe that the current Command interface is working well - right now we have several places that rely on the Args themselves to handle all sort of options. I don't think we have a need to differentiate them right now. I left a question for @fbrissi to understand why he chose to have that new interface/specific solution, instead of simply leveraging the `Command.AppendArgs`

danielsuguimoto:
@fabriciojs the profile option is not a `docker-compose up` argument...but it's a `docker-compose` one, it should be added before `up` command...I understood that @fbrissi wants to put the args before the command, instead of appending it after the command...

commands/start.go
fbrissi: Because the profile command is an argument to the docker-compose command, if I do whou you sugested, the result will be `docker-compose up --profile profile-1 -d` and this is an invalid command, the valid command is `docker-compose --profile profile-1 up -d`, so I created this new interface, which gives me access to the `localDockerCompose` property which is a command for `docker-compose`, that's where I append the args, because it would be a layer before star and I understand that it would be responsible for play the start before the up, even this attribute receives cmd up.

fbrissi: In issue #391 I send an image of the result of this code. But I couldn't create a test with this structure.

My first attempt was to add to attach, as you suggested, but i received invalid arguments from `docker-compose up`

fabriciojs (approved):

PR #398, opened 2021-12-08
upgrade to node 16

fabriciojs:
| Issue | - |
| -----: | :-----: |

**Description**

Updating all templates to ship with Node 16 image references instead of old Node 14.


templates/app/node14-adonis.yml
```diff
@@ -1,6 +1,6 @@
 services:
   app:
-    image: kooldev/node:14-adonis
+    image: kooldev/node:16-adonis
```
danielsuguimoto, line 3: we should rename this file to `node-adonis.yml` and change its references as well

templates/app/node14-nestjs.yml
```diff
@@ -1,6 +1,6 @@
 services:
   app:
-    image: kooldev/node:14-nest
+    image: kooldev/node:16-nest
```
danielsuguimoto, line 3: we should rename this file to `node-nestjs.yml` and change its references as well

danielsuguimoto (changes requested):
Nice! Minor comments about some files which name has the version 14 on it, but you changed the version to node 16. I think we should rename and make them more generic.

fabriciojs:
@danielsuguimoto good catch! Have updated the PR accordingly.

danielsuguimoto (approved):

PR #417, opened 2022-06-11
improve error message on script called without the 'run' command

fabriciojs:
| Issue | - |
| -----: | :-----: |
| :trophy: Feature | Yes |

**Description**

This PR generates a more helpful error message when trying to run what could be a script from `kool run foo` as simply `kool foo`.

### DIFF
--- a/commands/root.go
+++ b/commands/root.go
@@ -6,6 +6,7 @@ import (
 	"kool-dev/kool/core/environment"
 	"kool-dev/kool/core/parser"
 	"kool-dev/kool/core/shell"
+	"path"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


commands/root.go
```diff
@@ -72,6 +76,27 @@ Complete documentation is available at https://kool.dev/docs`,
...
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
+			if len(args) == 0 {
+				err = cmd.Help()
+				return
+			}
+
+			scriptParser := parser.NewParser()
+
+			// look for kool.yml on current working directory
+			_ = scriptParser.AddLookupPath(environment.NewEnvStorage().Get("PWD"))
```
danielsuguimoto, line 88: The `run` command also looks for scripts inside `$HOME/kool` folder
https://github.com/kool-dev/kool/blob/1ed97f479e099fb13f30dd5b3ba57b55f50b801b/commands/run.go#L69

danielsuguimoto (changes requested):

danielsuguimoto:
@fabriciojs added just one note. Other than that I think it's good to go.

Issue #418, opened 2022-06-13
Update images on Dockerfile


danielsuguimoto:
**Describe the feature**

- [x] Update docker-compose version on Dockerfile. Current: `1.28.2` Latest: `1.29.2`
- [x] Update alpine version on Dockerfile. Current: `3.12` Latest: `3.16`

fabriciojs:
This has been merged in and will be on the next release.

PR #421, opened 2022-07-01
Updating versions

fabriciojs:
| Issue | https://github.com/kool-dev/kool/issues/418 |
| -----: | :-----: |
| :warning: Break Change | TBD |

**Description**

Updating the docker-compose and base alpine image; the image is used on CI, would be important to test this before merging; but in case of problems one can always fix a past image while we work fixing moving forward, so I am OK adding this regardless of testing on CI.

---

**Notes**


fabriciojs:
OK alpine has a critical vulnerability in the latest image; we can hold on this PR while we get a patch in the official image possibly.

CVE-2022-28391  -  https://gitlab.alpinelinux.org/alpine/aports/-/issues/13661

fabriciojs:
Well actually it seems like 1.16 is not stable at; probably 1.15 is already patched; we can stick to that one.

danielsuguimoto (approved):
LGTM

PR #447, opened 2022-12-03
Compose v2

fabriciojs:
| Issue | - |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |
| :pencil: Refactor | Yes |

**Description**

Dropping `docker-compose` to always rely on `docker compose` - Compose V2.


commands/exec.go
```diff
@@ -54,11 +53,6 @@ func (e *KoolExec) detectTTY() {
 	if !isTerminal {
 		e.composeExec.AppendArgs("-T")
 	}
-
-	if aware, ok := e.composeExec.(compose.TtyAware); ok {
```
dbpolito, line 58: `docker compose` has better tty support / awareness?

core/shell/shell.go
```diff
@@ -144,7 +144,7 @@ func (s *DefaultShell) Exec(command builder.Command, extraArgs ...string) (outSt
 
 	cmd = execCmdFn(exe, args...)
 	cmd.Env = os.Environ()
-	cmd.Stdin = s.InStream()
+	// cmd.Stdin = s.InStream()
```
dbpolito, line 147: why?

dbpolito (approved):

dbpolito:
This PR is looking good, just a few questions to double check

commands/exec.go
fabriciojs: No, it's the same as it always was. What we removed is the double-trouble DinD option, where we had to set it twice.

core/shell/shell.go
fabriciojs: Good catch, this is a bad test change that we must undo.

Issue #528, opened 2025-10-27
Docker compose version output change


i-defranca:
**Describe the bug**

The install script checks the `docker compose version` output against the literal string *"Docker Compose version v2"* and as for now, at least on arch (EndeavourOS) I get the following output, which fails the check and stops the installation process: `Docker Compose version 2.*`

**Kool version and environment**

I'm using the current script available at https://kool.dev/install.

**To Reproduce**

Just follow the docs and try to install using curl | bash. Downloading the script first and removing the "v" from the string is enough for the it to run as expected.
 
**Screenshots**

<img width="827" height="234" alt="Image" src="https://github.com/user-attachments/assets/33610308-fc86-4dec-9c58-f8e1998d43b3" />

<img width="707" height="65" alt="Image" src="https://github.com/user-attachments/assets/1c16deca-9faa-4128-a5e3-621224cc767c" />

<img width="577" height="142" alt="Image" src="https://github.com/user-attachments/assets/281a9f8a-27c6-45b2-80ba-1e811f8d6abf" />

**Additional context**

I don't know in which version the output has changed, my current Docker Compose version is 2.40.2

FelipeEstevanatto:
Apparently, the version check in some distros can come up without the v in `v2`
I also checked on my machine because I have `Docker Compose version v5.1.0`, and the bash install also fails in the same manner, even with everything working fine in the newer version.
To fix these cases, the script would need to use a more permissive regex, like:
```bash
# ...
composeVersion=$(docker compose version || true)
regex="Docker Compose version v?([25]|[1-9][0-9]+)\."

if [[ ! "$composeVersion" =~ $regex ]]; then
	builtin echo -e "${start_error}We could not identify Composer V2 or V5 installed.${end_error}"
# ...
```

PR #532, opened 2026-04-02
fix: replace string match with regex for Compose version

i-defranca:
| Issue | #528 |
| -----: | :-----: |
| :beetle: Bug Fix | Yes |

**Description**
---
Use a regex-based check for Docker Compose version detection in the installation script, making the 'v' prefix optional and allowing versions 2, 5, and 10+.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -85,11 +85,13 @@ do_install () {
 	fi
 
 	composeVersion=$(docker compose version || true)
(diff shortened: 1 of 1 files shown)


fabriciojs:
Thank you @i-defranca - we just got a fix for this in now, didn't get you had already submitted one.

